PR #557, opened 2021-03-03
fix events of github flow

boscohyun:
As this repository became public, there was a problem that the `Repository Secrets` could not be accessed by `Pull Request` made based on the forked repository.
As a workaround, we considered using `Organization Secrets` and using `pull_request_target` event instead of `pull_request` event, and chose the latter.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,14 +12,14 @@ jobs:
     steps:
       # 클론
       - uses: actions/checkout@v2
-        if: github.event_name != 'pull_request'
+        if: github.event_name != 'pull_request_target'
         with:
           submodules: false
           lfs: false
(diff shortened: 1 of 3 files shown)


area363 (approved):

boscohyun:
Close this pull request.

Warning: The pull_request_target event is granted a read/write repository token and can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch, and to help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered.
From [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target).

PR #581, opened 2021-03-12
feat: do not run headless before snapshot finished

moreal:
It makes the launcher not start the headless before snapshot downloading ended up.

- [x] Use `chain tip` subcommand to check it should download snapshot or not.
- [x] Use `key *` subcommand to fetch and handle key store.
- [x] Use `validation private-key` subcommand to validate private-key.
- [x] Use `key convert --address [PRIVATE-KEY]` to convert private key into address format.

See also https://github.com/planetarium/NineChronicles.Headless/issues/342, https://github.com/planetarium/NineChronicles.Headless/issues/355, https://github.com/planetarium/NineChronicles.Headless/pull/367


moreal:
I opened this pull reqeust though the build failed, because it fails by `Lib9c` hash difference between `NineChronicles` and `NineChronicles.Headless`.

ipdae:
check-submodule is failed.

src/renderer/views/login/MiningView.tsx
```diff
@@ -15,6 +15,7 @@ const MiningView = observer(
     const classes = miningViewStyle();
     const setMining = (isMining: boolean) => {
       routerStore.push("/lobby/preload");
+      console.log("set-private-key setMining", accountStore.privateKey);
```
longfin, line 18: I guess it should be removed ㊙️

limebell (approved):
👍

area363 (approved):

src/renderer/views/login/MiningView.tsx
moreal: I removed it in 1bde93b

PR #586, opened 2021-03-16
Add some exceptions and handle it

boscohyun:
SSIA

image

image

image


src/main/main.ts
```diff
@@ -673,10 +685,24 @@ async function initializeStandalone(): Promise<void> {
 
           break;
         } catch (error) {
-          console.error(
-            `Unexpected error occurred during download / extract snapshot. ${error}` +
-              `path: ${path}`
-          );
+          errorOccurred = true;
+          const errorMessage = `Unexpected error occurred during download / extract snapshot.\n${error}`;
```
ipdae, line 689: `path` is deleted. is this correct change?

src/main/main.ts
```diff
@@ -613,6 +624,7 @@ async function initializeStandalone(): Promise<void> {
   }
 
   initializeStandaloneCts = CancellationToken.create();
+  let errorOccurred = false;
```
ipdae, line 627: there are no use of this variable.

src/utils.ts
```diff
@@ -144,6 +151,10 @@ export async function cancellableExtract(
...
       `Unexpected error occurred during extracting ${targetDir} to ${outputDir}. ${error}`
     );
-    throw error;
+    throw new CancellableExtractFailedError(targetDir, outputDir);
   }
 }
+
+export function getType(target: any) {
+  return Object.prototype.toString.call(target).slice(8, -1);
+}
```
ipdae, line 160:
```suggestion
}

```

src/main/main.ts
boscohyun: Yes it is. The `path` used as argument of `snapshot.downloadMetadata()` and `snapshot.downloadSnapshot()`. And that used to make the `downloadPath` variable inside of these methods. Each methods throw new exceptions now(`DownloadSnapshotFailedError`, `DownloadSnapshotMetadataFailedError`) and these exceptions error messages contains the `downloadPath`.

riemannulus (approved):

area363 (approved):

PR #588, opened 2021-03-17
Wait for synchronously set private key ended

ipdae:
- fix https://github.com/planetarium/NineChronicles.Headless/issues/375

### DIFF
--- a/src/renderer/views/login/MiningView.tsx
+++ b/src/renderer/views/login/MiningView.tsx
@@ -15,15 +15,15 @@ const MiningView = observer(
     const classes = miningViewStyle();
     const setMining = (isMining: boolean) => {
       routerStore.push("/lobby/preload");
-      if (
-        !ipcRenderer.sendSync(
-          "standalone/set-private-key",
-          accountStore.privateKey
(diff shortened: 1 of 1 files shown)


src/renderer/views/login/MiningView.tsx
```diff
@@ -15,15 +15,15 @@ const MiningView = observer(
...
     const setMining = (isMining: boolean) => {
       routerStore.push("/lobby/preload");
-      if (
-        !ipcRenderer.sendSync(
-          "standalone/set-private-key",
-          accountStore.privateKey
-        )
-      ) {
+      const setSucceed: boolean = ipcRenderer.sendSync(
+        "standAlone/set-private-key",
```
moreal, line 19: It seems there is no changes about chanel renaming but it was changed into `standAlone`.

https://github.com/planetarium/9c-launcher/blob/ca76d73a0acaf55effad77cfed28704f7f53a939/src/main/standalone.ts#L39

Is it intended?

moreal (approved):

area363 (approved):

PR #602, opened 2021-03-24
fix snapshot download fallback

riemannulus:
Fixed an issue where if any one of the snapshot paths failed, the launcher would go to the error page, and changed it so that if all of the snapshot paths failed, headless would not appear. 

plus, change name from standalone to headless


src/main/main.ts
```diff
@@ -667,27 +663,34 @@ async function initializeStandalone(isRestart: boolean = false): Promise<void> {
       );
     } else if (snapshotPaths.length > 0 && win != null) {
       const snapshotDownloadUrls: string[] = electronStore.get("SnapshotPaths");
-      if (snapshotDownloadUrls.length > 0 && win != null && !isRestart) {
```
longfin, line 670: Is it okay that we ignore `isRestart` here?

src/main/main.ts
```diff
@@ -667,27 +663,34 @@ async function initializeStandalone(isRestart: boolean = false): Promise<void> {
...
-            app.getPath("userData"),
-            standalone,
-            win,
-            initializeStandaloneCts.token
-          );
-          if (isProcessSuccess) break;
-        }
+      let isProcessSuccess = false;
+      for (const snapshotDownloadUrl of snapshotDownloadUrls) {
+        isProcessSuccess = await snapshot.processSnapshot(
```
longfin, line 668: it can't be a big deal... but IMHO, handling an exception from `processSnapshot()` can be better than using boolean return value because we can get more detailed information about the exception.

src/errors/StandaloneInitializeError.ts
```diff
@@ -1,8 +1,8 @@
-class StandaloneInitializeError extends Error {
+class HeadlessInitializeError extends Error {
```
boscohyun, line 1: How about changing the file name too.

boscohyun (approved):

src/main/main.ts
riemannulus: Yes. The reason `isRestart` was used was because of the old partition snapshot implementation and is no longer needed.

src/main/main.ts
riemannulus: Ok, I'll check try-catch on `main` not `snapshot`

riemannulus: 7585794 changed

longfin (approved):

area363 (approved):

PR #604, opened 2021-03-25
Relaunch headless when it exited with unexpected error

limebell:
This patch let headless to relaunch when headless is exited with unexpected error. Additionally it prevents duplicated initialize process.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -626,6 +626,20 @@ async function initializeHeadless(): Promise<void> {
   */
   console.log(`Initialize headless. (win: ${win?.getTitle})`);
 
+  if (initializeHeadlessCts !== null) {
+    console.error(
+      "Cannot initialize headless while initializing headless.",
+    );
+    return;
+  }
+
+  if (standalone.alive) {
+    console.error(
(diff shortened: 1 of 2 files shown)


src/main/main.ts
```diff
@@ -627,6 +626,20 @@ async function initializeHeadless(): Promise<void> {
...
   console.log(`Initialize headless. (win: ${win?.getTitle})`);
 
+  if (initializeHeadlessCts !== null) {
+    console.error(
+      "Cannot initialize headless while initializing headless.",
+    );
+    return;
+  }
+
+  if (standalone.alive) {
```
riemannulus, line 636: Can you change `standalone` to `headless`?

riemannulus (commented):
I think we can be targeted when kill by RPC error, (ex, error code 3762504530). How about that?

src/main/main.ts
limebell: It would be large diff including file name change, I guess it is better to be in another PR.

riemannulus (approved):

area363 (approved):

PR #605, opened 2021-03-25
Use DefinitelyTyped definitions instead own type definition

x86chi:
When I create own type definitions for this project, `@types/zxcbvn` was in NPM. 

But there was no have literal type about password hint like below:
```ts
type ZXCVBNFeedbackWarning =
  | "This is a top-10 common password"
  | "This is a top-100 common password" ...
```

Recently, I had update `zxcvbn` definition to DefinitelyTyped:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51688

Also, I create `@types/textfit` definition:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51809


package-lock.json
```diff
@@ -3878,6 +3881,15 @@
...
     },
+    "node_modules/@types/axios": {
+      "version": "0.14.0",
+      "resolved": "https://registry.npmjs.org/@types/axios/-/axios-0.14.0.tgz",
+      "integrity": "sha1-7CMA++fX3d1+udOr+HmZlkyvzkY=",
+      "deprecated": "This is a stub types definition for axios (https://github.com/mzabriskie/axios). axios provides its own type definitions, so you don't need @types/axios installed!",
+      "dependencies": {
+        "axios": "*"
+      }
+    },
```
x86chi, line 3892: I don't know why add `@types/axios` to `package-lock.json`. 🤔

moreal: [`@types/axios` has been in `package.json`](https://github.com/planetarium/9c-launcher/blob/1b8acf4988563b639aa9b886f101804258666be1/package.json#L62) but the `package-lock.json` seems not updated at that time. And `@types/axios` should be removed because `@types/axios` was deprecated and `axios` has supported its type itself.

For `node_modules/@types/axios`, According to [GitHub (npm) blog, there was a patch related to *package-lock.json* since npm v7](https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/). So IMO it seems you are using npm v7 and the `node_modules/@types/axios` were added also.

moreal (commented):
👍🏻 Could you rebase this pull request branch on the target branch?

package-lock.json
x86chi: It was added at 52a215d3d3e1dd340e033da6ef0d875a1a927c61. @riemannulus is it intended change? I think is not related to fixing tsconfig path.

riemannulus: That's not intended. It is a just mistake.

moreal (approved):
LGTM 👍🏻

package-lock.json
x86chi: Can I remove `@types/axios` in this PR? (or create new PR?)

moreal: In my thought, it is better to remove in new PR.

longfin (approved):

PR #609, opened 2021-03-29
feat(graphql): use session

moreal:
This pull request makes the launcher client to call `mutation.auth.login` while sharing session between `main` and `renderer` process.

### DIFF
--- a/src/main/headless.ts
+++ b/src/main/headless.ts
@@ -1,5 +1,5 @@
 import { ChildProcess, execFileSync } from "child_process";
-import { ipcMain } from "electron";
+import { ipcMain, net, Session, session, webContents } from "electron";
 import { dirname, basename } from "path";
 import { electronStore, CUSTOM_SERVER, LOCAL_SERVER_URL } from "../config";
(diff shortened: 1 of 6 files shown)


moreal:
Could you review this pull request, reviewers?

src/main/headless.ts
```diff
@@ -120,7 +126,19 @@ class Headless {
     const body = JSON.stringify({
       PrivateKeyString: privateKey,
     });
-    return this.retriableFetch("set-private-key", body);
+    const httpResult = await this.retriableFetch("set-private-key", body);
```
longfin, line 129: Just curious, do we still need `set-private-key` endpoint?

longfin (approved):

src/main/headless.ts
moreal: At this time, yes. It is for mining. To remove it, https://github.com/planetarium/NineChronicles.Headless/pull/413 pull request should be merged.

longfin: 👌

area363 (approved):

PR #619, opened 2021-03-31
Fix bug in copying private key

boscohyun:
SSIA

### DIFF
--- a/src/errors/UndefinedProtectedPrivateKeyError.ts
+++ b/src/errors/UndefinedProtectedPrivateKeyError.ts
@@ -1,0 +1,8 @@
+class UndefinedProtectedPrivateKeyError extends Error {
+  constructor(msg: string) {
+    super(msg);
+    Object.setPrototypeOf(this, UndefinedProtectedPrivateKeyError.prototype);
+  }
+}
+
+export default UndefinedProtectedPrivateKeyError;
--- a/src/errors/UndefinedProtectedPrivateKeyError.ts
+++ b/src/errors/UndefinedProtectedPrivateKeyError.ts
@@ -1,7 +1,6 @@
 class UndefinedProtectedPrivateKeyError extends Error {
   constructor(msg: string) {
     super(msg);
-    Object.setPrototypeOf(this, UndefinedProtectedPrivateKeyError.prototype);
   }
 }
 
--- a/src/renderer/views/account/CreateAccountView.tsx
+++ b/src/renderer/views/account/CreateAccountView.tsx
@@ -34,11 +34,20 @@ const CreateAccountView = observer(
         password
       );
 
-      const privateKey: PrivateKey = ipcRenderer.sendSync(
-        "unprotect-private-key",
-        address,
-        password
-      );
+      const unprotectedPrivateKey: [
+        string | undefined,
+        any
+      ] = ipcRenderer.sendSync("unprotect-private-key", address, password);
+      if (
+        typeof unprotectedPrivateKey[0] !== "string" ||
+        unprotectedPrivateKey[0] === ""
+      ) {
+        // FIXME: Show a new error page or retry page to complete the account creation.
+        console.error("Failed to unprotect private key");
+        return;
+      }
+
+      const privateKey: string = unprotectedPrivateKey[0];
 
       accountStore.setPrivateKey(privateKey);
       accountStore.addAddress(address);
--- a/src/renderer/views/account/CreateAccountView.tsx
+++ b/src/renderer/views/account/CreateAccountView.tsx
@@ -36,14 +36,14 @@ const CreateAccountView = observer(
 
       const [privateKey, error]: [
         string | undefined,
-        any | undefined
+        Error | undefined
       ] = ipcRenderer.sendSync("unprotect-private-key", address, password);
       if (
         error !== undefined ||
         privateKey === undefined ||
         privateKey === "") {
         // FIXME: Show a new error page or retry page to complete the account creation.
-        console.error("Failed to unprotect private key");
+        console.error(`Failed to unprotect private key. ${error?.name}: ${error?.message}`);
         return;
       }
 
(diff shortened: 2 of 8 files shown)


src/renderer/views/account/CreateAccountView.tsx
```diff
@@ -34,11 +34,20 @@ const CreateAccountView = observer(
...
+      if (
+        typeof unprotectedPrivateKey[0] !== "string" ||
+        unprotectedPrivateKey[0] === ""
+      ) {
+        // FIXME: Show a new error page or retry page to complete the account creation.
+        console.error("Failed to unprotect private key");
+        return;
+      }
+
+      const privateKey: string = unprotectedPrivateKey[0];
```
moreal, line 50: How about using [`Destructuring assignment`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)? And it passes also error object like Go does. So how about checking error with it? Like...

```suggestion
      const [privateKey, error]: [
        string | undefined,
        Error | undefined
      ] = ipcRenderer.sendSync("unprotect-private-key", address, password);
      if (
        error !== undefinded ||
        typeof unprotectedPrivateKey[0] !== "string" ||
        unprotectedPrivateKey[0] === ""
      ) {
        // FIXME: Show a new error page or retry page to complete the account creation.
        console.error("Failed to unprotect private key");
        return;
      }
```

area363 (approved):

src/renderer/views/account/CreateAccountView.tsx
boscohyun: @moreal I fixed it!

src/renderer/views/account/CreateAccountView.tsx
```diff
@@ -34,11 +34,15 @@ const CreateAccountView = observer(
         password
       );
 
-      const privateKey: PrivateKey = ipcRenderer.sendSync(
-        "unprotect-private-key",
-        address,
-        password
-      );
+      const [privateKey, passphraseOrError]: [
```
moreal, line 37: It seems the second element of the return value has been never `passphrase`.

https://github.com/planetarium/9c-launcher/blob/78c0ee02eca6a18df6af74d1dcee1b0cf3979bb4/src/main/main.ts#L539-L562

So it should rename into `error`.

boscohyun: Oh my mistake. The js(ts) convention is confusing to me.😢

boscohyun: The second element can be `{}`, `undefined`, `error: any`. How can I naming it? Hm... Just `any` or `_` I think.

boscohyun: Here is [a interesting article](https://ntgard.medium.com/tuples-in-javascript-cd33321e5277)
```
const [, , thisOne, , thatOne] = getTuple();
const [_1, _2, thisOne, _3, thatOne] = getTuple();
```

moreal: Then, I suggest to ignore the second element like the post you found `const [privateKey] = ipc...` if you will check the error with only `privateKey` state without `error`. But I suggest also to check with `error`.

And I have a question, what the case the second element became `{}`?

boscohyun: The second element became `{}` on line 547.

moreal (approved):

src/renderer/views/account/CreateAccountView.tsx
moreal: I see 👍🏻

boscohyun: The `{}` on line 547 can change to `error: string` that contains error message. How's it? @moreal 
And store the second element as `error: any | undefined` type and handles the error.

moreal: > And store the second element as `error: any | undefined` type and handles the error.

`any` contains `undefined` so there is no reason to use `any | undefined`.

> The `{}` on line 547 can change to `error: string` that contains error message

Could you describe more about the change you mean? If you mean making it return other meaningful value instead `{}`, it seems better to return `Error` instead `string`, to suit for `Error` concepts.

src/errors/UndefinedProtectedPrivateKeyError.ts
```diff
@@ -0,0 +1,8 @@
+class UndefinedProtectedPrivateKeyError extends Error {
+  constructor(msg: string) {
+    super(msg);
+    Object.setPrototypeOf(this, UndefinedProtectedPrivateKeyError.prototype);
+  }
+}
+
+export default UndefinedProtectedPrivateKeyError;
```
riemannulus, line 8: Can we use function component instead class component? https://ko.reactjs.org/docs/components-and-props.html#function-and-class-components

moreal: It seems an inheritance of `Error` class, not `React.Component`.

riemannulus: Oh yes 😅  I misreading class name lol

riemannulus (approved):

src/errors/UndefinedProtectedPrivateKeyError.ts
```diff
@@ -0,0 +1,8 @@
+class UndefinedProtectedPrivateKeyError extends Error {
+  constructor(msg: string) {
+    super(msg);
+    Object.setPrototypeOf(this, UndefinedProtectedPrivateKeyError.prototype);
```
moreal, line 4: It is a question, why does it call [`Object.setPrototypeOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf)? (P.S. I read [Mozilla Custom Error Types section](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#custom_error_types).)

boscohyun: I'm curious too.🤔

boscohyun: I agree.

PR #628, opened 2021-04-06
Fix preloading message bug

boscohyun:
Set progressMessage in useEffect method for isPreloadEnded, step, progress.


src/renderer/views/preload/PreloadProgressView.tsx
```diff
@@ -29,7 +28,7 @@ const PreloadProgressView = observer(() => {
   } = useNodeExceptionSubscription();
   const preloadProgress = preloadProgressSubscriptionResult?.preloadProgress;
 
-  const [isPreloadEnded, setPreloadStats] = useState(false);
+  const [isPreloadEnded, setIsPreloadEnded] = useState(false);
```
longfin, line 31: It isn't a big deal, but IMHO `preloadEnded` and `setPreloadEnded` are more natural for me.

longfin (approved):

PR #631, opened 2021-04-07
feat: do not call `/run-standalone` endpoint

moreal:
Since https://github.com/planetarium/NineChronicles.Headless/pull/420, `--libplanet-node` option was deprecated and it became to start libplanet headless as default. so it doesn't need to call `/run-standalone` and I applied the change in this pull request.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -41,7 +41,11 @@ import * as utils from "../utils";
 import * as partitionSnapshot from "./snapshot";
 import * as monoSnapshot from "./monosnapshot";
 import Headless from "./headless";
-import { HeadlessExitedError, HeadlessInitializeError, UndefinedProtectedPrivateKeyError } from "../errors";
+import {
+  HeadlessExitedError,
+  HeadlessInitializeError,
+  UndefinedProtectedPrivateKeyError,
+} from "../errors";
 import CancellationToken from "cancellationtoken";
 import { IDownloadProgress, IGameStartOptions } from "../interfaces/ipc";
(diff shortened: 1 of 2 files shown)


src/main/main.ts
```diff
@@ -81,6 +85,7 @@ const standaloneExecutableArgs = [
       (trustedAppProtocolVersionSigner) =>
         `-T=${trustedAppProtocolVersionSigner}`
     ),
+  "--no-miner",
```
longfin, line 88: We may need to set this option as default on the headless side🤔 ...

longfin (approved):

longfin:
Lib9cs in submodules(NineChronicles, NineChronicles.Headless) aren't the same... could you bump it? @moreal

dahlia (approved):

src/main/main.ts
moreal: I opened the bump pull request, https://github.com/planetarium/NineChronicles/pull/95. After it was merged, I will update this pull request 🙏🏻

moreal:
> 
> 
> Lib9cs in submodules(NineChronicles, NineChronicles.Headless) aren't the same... could you bump it? @moreal

I opened the bump pull request, https://github.com/planetarium/NineChronicles/pull/95. After it was merged, I will update this pull request 🙏🏻

limebell (approved):

moreal:
I rebased and bumped submodules but there is an issue which the `Pack` worfklow was broken, https://github.com/planetarium/NineChronicles/issues/108.

longfin (approved):

PR #643, opened 2021-04-12
feat: pass stdin when executing subcommand

moreal:
This pull request makes `StandaloneSubcommand.execSync` to support `stdin` input.

### DIFF
--- a/src/main/headless/validation.ts
+++ b/src/main/headless/validation.ts
@@ -4,7 +4,7 @@ import { StandaloneSubcommand } from "./subcommand";
 export class Validation extends StandaloneSubcommand {
   public isValidPrivateKey(privateKeyHex: string): boolean {
     try {
-      this.execSync("validation", "private-key", privateKeyHex);
+      this.execSync(["validation", "private-key", privateKeyHex]);
       return true;
     } catch (error) {
       console.error(error);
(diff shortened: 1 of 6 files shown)


src/main/headless/validation.ts
```diff
@@ -4,7 +4,7 @@ import { StandaloneSubcommand } from "./subcommand";
 export class Validation extends StandaloneSubcommand {
   public isValidPrivateKey(privateKeyHex: string): boolean {
     try {
-      this.execSync("validation", "private-key", privateKeyHex);
+      this.execSync(["validation", "private-key", privateKeyHex]);
```
longfin, line 7: Is it intended?

```suggestion
      this.execSync(["validation", "private-key"], privateKeyHex);
```

moreal: Yes, it was intended but the code you suggested is correct and right. But `NineChronicles.Headless.Executable validation private-key` doesn't support stdin check. It should be implemented in **planetairum/NineChronicles.Headless** first.

area363 (approved):
Branch update required!

moreal:
> 
> 
> Branch update required!

@area363 I updated this branch on the latest of the target branch and I updated one more file because the `src/main/headless/apv` command was introduced. Could you review it also please? 🙏🏻

area363:
E2E-Test seems to fail 😢

PR #685, opened 2021-05-20
style: fix awkward background image

riemannulus:
it is based at border frame. this pr add frame at landing and main page


src/collection/pages/main/loading.stories.tsx
```diff
@@ -0,0 +1,13 @@
...
+import Loading from "./loading";
+
+export default {
+  title: "Collection/Pages/Loading",
+  component: Loading,
+} as Meta;
+
+const Template: Story = () => <Loading />;
+
+export const Primary = Template.bind({});
```
boscohyun, line 13:
```suggestion
export const Primary = Template.bind({});

```

boscohyun (approved):

src/collection/pages/landing/landing.tsx
```diff
@@ -16,6 +16,7 @@ const Landing: React.FC<Props> = (props: Props) => {
       className={"IntroContainer"}
       style={{ backgroundImage: `url(${BackgroundImage})` }}
     >
+      <div className={'IntroPageBorder'}>
       <div className={"IntroItems"}>
       <div className={"Title"}>
```
longfin, line 21: IMO, it would be better if we having indentations 😅

src/collection/pages/main/loading.tsx
```diff
@@ -3,7 +3,7 @@ import React from "react"
 import './loading.scss';
 
 const LoadingPage: React.FC = () => {
-    return <div className={'LoadingPageContainer'}></div>
+    return <div className={'LoadingPageContainer'}><div className={'LoadingPageBorder'}></div></div>
```
longfin, line 6: Just curiosity, why we need `{}` instead of the plain attribute as below?
```suggestion
    return <div className="LoadingPageContainer"><div className="LoadingPageBorder"></div></div>
```

longfin (approved):

src/collection/pages/main/loading.tsx
riemannulus: It's just inertia 😅 `{}` is doesn't need in here

PR #686, opened 2021-05-20
fix: edge case for minerAddress

riemannulus:
Resolves an issue where the screen appears even though the loading is not finished.

### DIFF
--- a/src/collection/pages/main/main.tsx
+++ b/src/collection/pages/main/main.tsx
@@ -54,7 +54,7 @@ const Main: React.FC = () => {
       address: agentAddress,
     },
   });
-  const { data: minerAddress } = useMinerAddressQuery();
+  const { data: minerAddress, loading: minerAddressLoading } = useMinerAddressQuery();
   const [
     collect,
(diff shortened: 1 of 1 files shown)


boscohyun (approved):

longfin (approved):

src/collection/pages/main/main.tsx
```diff
@@ -161,11 +161,14 @@ const Main: React.FC = () => {
...
   }, [collectionState, data])
 
-  if (loading) return <LoadingPage/>;
+  if (loading || minerAddressLoading) return <LoadingPage/>;
   if (error) return <p>error</p>;
   if(data?.stateQuery.agent == null) {
     <div>you need create avatar first</div>
   }
+  if(minerAddress?.minerAddress == null) {
+    <div>you need login first</div>
```
longfin, line 170: 1. Is it okay that we don't have `return`?
2. it would be better if we have `FIXME` like as...

```
```
```suggestion
    // FIXME we should translate this message.
    <div>you need login first</div>
```

PR #688, opened 2021-05-20
fix: return error page when it's wrong

riemannulus:
Return an error page in case of incorrect access

### DIFF
--- a/src/collection/pages/main/main.tsx
+++ b/src/collection/pages/main/main.tsx
@@ -164,9 +164,11 @@ const Main: React.FC = () => {
   if (loading || minerAddressLoading) return <LoadingPage/>;
   if (error) return <div><p>Error: </p>{JSON.stringify(error)}</div>;
   if(minerAddress?.minerAddress == null) {
+    // FIXME we should translate this message.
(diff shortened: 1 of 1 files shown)


boscohyun (approved):

src/collection/pages/main/main.tsx
```diff
@@ -162,12 +162,12 @@ const Main: React.FC = () => {
...
-  if(data?.stateQuery.agent == null) {
-    <div>you need create avatar first</div>
-  }
+  if (error) return <div><p>Error: </p>{JSON.stringify(error)}</div>;
   if(minerAddress?.minerAddress == null) {
-    <div>you need login first</div>
+    return <div>you need login first</div>
+  }
+  if(data?.stateQuery.agent == null) {
+    return <div>you need create avatar first</div>
```
longfin, line 170: How about adding FIXME?

```suggestion
    // FIXME we should translate this message.
    return <div>you need login first</div>
  }
  if(data?.stateQuery.agent == null) {
    // FIXME we should translate this message.
    return <div>you need create avatar first</div>
```

longfin (approved):

PR #693, opened 2021-05-20
fix: remain message when do not collect

riemannulus:
Fix remain message when do not collect


src/collection/components/RemainingDisplay/RemainingDisplay.tsx
```diff
@@ -15,11 +16,11 @@ const getRemain = (remainHour: number) => {
 
   if(hour >= 1) return `${Math.round(hour)} hours`
 
-  return `less then hour`
+  return `less then 1 hour`
```
boscohyun, line 19:
```suggestion
  return `less then an hour`
```

boscohyun (approved):

PR #697, opened 2021-05-21
feat: logging about downloadsnapshot failed

riemannulus:


boscohyun (approved):

src/main/snapshot.ts
```diff
@@ -191,6 +191,7 @@ export async function downloadSnapshot(
     if (token.reason === "clear-cache") {
       throw new ClearCacheException();
     } else {
+      console.log(JSON.stringify(error));
```
longfin, line 194: Is `error`  encoded as JSON?

riemannulus: Yes. [(ref)](https://stackoverflow.com/a/18391400)

riemannulus: fixed: e353296

riemannulus:
PTAL @longfin

longfin (approved):

PR #707, opened 2021-05-24
Display floating notice when try remove 1 tier monster

riemannulus:
image

### DIFF
--- a/src/collection/components/Cart/Cart.tsx
+++ b/src/collection/components/Cart/Cart.tsx
@@ -7,6 +7,7 @@ import './Cart.scss';
 import CartItem from "./CartItem/CartItem";
 import CollectionButton from "../Button/Button";
 import stepIcon from "../../common/resources/bg-staking-slot-step.png";
+import Animated from "react-mount-animation";
 
 export type Props = {
     cartList: CollectionItemModel[],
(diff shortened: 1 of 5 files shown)


src/collection/components/Cart/Cart.tsx
```diff
@@ -27,12 +28,43 @@ const Cart: React.FC<Props> = (props: Props) => {
...
+      if(item.tier !== CollectionItemTier.TIER1) return;
+      
+      const latestItem = cartList.find(x => x.collectionPhase === CollectionPhase.LATEST);
+      
+      if(latestItem?.tier !== CollectionItemTier.TIER1) return;
+
+      displayWarning();
+    }
+
+    const displayWarning = async () => {
```
longfin, line 42: Is there any reason that this method is `async`?

longfin (approved):

src/collection/components/Cart/Cart.tsx
riemannulus: IMO if this method is executed synchronously, it will cause performance degradation to the renderer process.

riemannulus: I want to run this on another task, not renderer.

longfin: As far I know, [`setTimeout()`](https://developer.mozilla.org/ko/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout) doesn't block. it just adds a timer to the main event thread.

src/collection/components/Cart/Cart.tsx
```diff
@@ -27,12 +28,43 @@ const Cart: React.FC<Props> = (props: Props) => {
...
+      setWarning(true);
+      let timer = setTimeout(() => setWarning(false), 5 * 1000);
+
+      // this will clear Timeout
+      // when component unmount like in willComponentUnmount
+      // and show will not change to true
+      return () => {
+        clearTimeout(timer);
+      };
```
longfin, line 51: Is this return value needed? it seems that there is no usage.

riemannulus: That means the timer cleared when the component unmounted.

riemannulus:
PTAL 🙏  @longfin

longfin (approved):

PR #718, opened 2021-05-24
Disable monster collection editing

longfin:
This PR blocks editing the monster collection that has been registered. 
And it's block empty-applying on monster collection cart too.

### DIFF
--- a/src/collection/pages/main/main.tsx
+++ b/src/collection/pages/main/main.tsx
@@ -64,12 +64,12 @@ const Main: React.FC = () => {
     cancelCollection,
   ] = useCancelCollectionMutation();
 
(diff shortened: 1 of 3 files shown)


src/collection/pages/main/main.tsx
```diff
@@ -248,7 +248,7 @@ const Main: React.FC = () => {
       (x) => x.collectionPhase === CollectionPhase.LATEST
     );
 
-    if(!latestCollectionItem) return;
+    if (!latestCollectionItem) throw Error("Mutation target cannot be 0");
```
riemannulus, line 251: is it intented?

longfin: It isn't intended. 😢 I'll open another PR to address it.

PR #764, opened 2021-06-15
Replace existing localization system with Transifex Native

BasixKOR:
This PR aims to replace all the existing useLocale calls to appropriate Transifex Native API calls. However there are some pitfalls, mainly multiple string entries being replaced to simple multiline strings.

Note that this PR is pretty much WIP, and there are major issues to be resolved before merging this into upstream.

- [x] Connect Transifex API keys
- [x] Automatic string update, preferably with GitHub Actions
- [ ] ~~Find a way to use current API with Monster Collections system.~~ Postponed.


BasixKOR:
Oh right, before merging this please make sure to put Transifex API keys to GitHub secrets. Without it the action will not be able to update translations!

src/renderer/views/error/ErrorDiskSpaceView.tsx
```diff
@@ -4,29 +4,26 @@ import { Typography } from "@material-ui/core";
...
   }, []);
   return (
     <div className={classes.root}>
       <Typography variant="h1" gutterBottom className={classes.title}>
-        {locale("디스크 공간이 충분하지 않습니다")}
+        <T _str="Disk space is not enough." _tags="errorDiskSpace" />
       </Typography>
       <Typography>
-        {locale("필요한 여유 공간:")} {prettyBytes(REQUIRED_DISK_SPACE)}
+        <T _str="Required free space:" _tags="errorDiskSpace" /> {prettyBytes(REQUIRED_DISK_SPACE)}
```
longfin, line 23: IMO, we can use like format string.

```suggestion
        <T _str="Required free space: {size}" size={prettyBytes(REQUIRED_DISK_SPACE)| _tags="errorDiskSpace" />
```

BasixKOR: Nice suggestion! Should we apply that to other string in this file as well?

src/renderer/components/RetypePasswordForm.tsx
```diff
@@ -7,16 +7,76 @@ import {
 } from "@material-ui/core";
 import { makeStyles } from "@material-ui/styles";
 import React, { useState, ChangeEvent, MouseEvent } from "react";
-import zxcvbn from "zxcvbn";
+import zxcvbn, { ZXCVBNFeedbackWarning } from "zxcvbn";
 import { RetypePassword } from "../../interfaces/i18n";
-import { useLocale } from "../i18n";
+import { T } from "@transifex/react";
+import { t } from "@transifex/native";
```
riemannulus, line 13: What's different between `t` and `T`?

src/renderer/components/RetypePasswordForm.tsx
```diff
@@ -162,29 +220,31 @@ const RetypePasswordForm = ({
...
+            <T _str="Invitation Code" _tags="retypePassword" />
+          </InputLabel>
+          <OutlinedInput type="text" onChange={handleActivationKeyChange} />
+          <FormHelperText className={classes.helperText}>
+            form helper text
+          </FormHelperText>
+        </FormControl>
+      ) : (
+        <></>
+      )}
```
riemannulus, line 239: Maybe we can use `&&` condition render instead `?`

```suggestion
      {useActivationKey && (
        <FormControl
          fullWidth
          error={isActivationKeyError}
          className={classes.formControl}
        >
          <InputLabel className={classes.label}>
            <T _str="Invitation Code" _tags="retypePassword" />
          </InputLabel>
          <OutlinedInput type="text" onChange={handleActivationKeyChange} />
          <FormHelperText className={classes.helperText}>
            form helper text
          </FormHelperText>
        </FormControl>
      )}
```

src/renderer/stores/game.ts
```diff
@@ -22,10 +21,6 @@ export default class GameStore {
     this._appProtocolVersion = electronStore.get(
       "AppProtocolVersion"
     ) as string;
-
-    if (!(this._language in supportedLocales)) {
-      this._language = "en";
-    }
```
riemannulus, line 28: We need to leave this option for Unity Game. Can't we check the supported locale from Transifex?

src/renderer/views/account/AccountView.tsx
```diff
@@ -5,41 +5,40 @@ import { Container, InputLabel } from "@material-ui/core";
...
-    const { locale } = useLocale<Account>("account");
     return (
       <div>
         <button
           onClick={() => {
             routerStore.push("/account/create");
           }}
         >
-          {locale("키 생성하기")}
+          <T _str="create key" _tags="account" />
```
riemannulus, line 20: IMO use variables instead of a scala-type string are better.

src/renderer/components/RetypePasswordForm.tsx
BasixKOR: `T` is a React component, and `t` is the underlying function for retrieving translated strings.

src/renderer/stores/game.ts
BasixKOR: I haven't found any validation functionality provided by Transifex. I'll check if any other third-party libraries can help.

I doubt that this should be done in simple equality check. For instance, if the user locale is `en_GB` it won't be equal to `en` or `en_US`. Although this exact case won't be the problem since English is the default.

src/renderer/views/account/AccountView.tsx
BasixKOR: You mean for every tag initialization?

src/renderer/components/RetypePasswordForm.tsx
BasixKOR: Uh, actually I think we gotta avoid doing that. See: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

src/renderer/views/account/AccountView.tsx
riemannulus: Yes, like this
```suggestion
          <T _str="create key" _tags=`${transifexTag}` />
```

src/renderer/stores/game.ts
riemannulus: Yes, but at least I knew, the game needs default language and it's only `en`, do not need detail like `en_US`.

src/renderer/components/RetypePasswordForm.tsx
riemannulus: Could we use only one type?

src/renderer/views/account/AccountView.tsx
BasixKOR: Seems like nice idea to me! But Transifex CLI's const lookup doesn't work on some APIs, so I filed a PR to resolve this issue. Hopefully they'll merge it and we could use them.

src/renderer/stores/game.ts
BasixKOR: What I meant is some users may experience fallback to English even if their language is supported on the game. Maybe we could address this one in detail in next meetup.

src/renderer/components/RetypePasswordForm.tsx
BasixKOR: I thought that would be impossible but after looking the code back, I think I could refactor to using a React variant only.

src/renderer/components/RetypePasswordForm.tsx
x86chi: How about use `!!useActivationKey &&` ?

src/renderer/stores/game.ts
BasixKOR: The validation step is moved to the provider's effects, and the GameStore should be able to update its language by the added snippet subscribing to locale changes.

riemannulus (approved):

riemannulus:
PTAL @longfin 🙏

longfin (approved):

PR #785, opened 2021-06-25
fix(i18n): Fix various bugs

BasixKOR:
* Prevent current locale to be overwritten to English, the fallback language.
* Prevent Select from using intermediate value as default.
* Fixes to properly fetch translations after initialization (upstream bug).

### DIFF
--- a/src/renderer/views/config/ConfigurationView.tsx
+++ b/src/renderer/views/config/ConfigurationView.tsx
@@ -1,7 +1,7 @@
 import path from "path";
 
 import { ipcRenderer, remote, shell } from "electron";
-import React from "react";
+import React, { useMemo } from "react";
 import { observer } from "mobx-react";
 import {
   Button,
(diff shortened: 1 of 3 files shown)


src/renderer/views/config/ConfigurationView.tsx
```diff
@@ -116,19 +124,18 @@ const ConfigurationView = observer(() => {
...
-            }
-          />
+          {languages.length > 0 && (
+            <Select
+              name="select"
+              className={classes.selectLocale}
+              items={languages.map(({ localized_name }) => localized_name)}
+              defaultValue={selectedLanguage ?? "English"}
+            />
+          )}
```
riemannulus, line 136: If `lanugages` length is 0, the configuration form is not broken? IMO the css style is might be little different.

BasixKOR: It seems almost okay with just the Select missing while it's fetching the languages, but it might be good idea to add indicator like a loading spinner.

riemannulus (approved):

PR #786, opened 2021-06-28
Send mixpanel event when relanch headless

riemannulus:
SSIA

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -544,7 +544,12 @@ function initializeIpc() {
     }
   });
 
-  ipcMain.on("relaunch standalone", async (event) => {
+  ipcMain.on("relaunch standalone", async (event, param: {}) => {
+    mixpanel?.track("Launcher/Relaunch Headless", {
+      distinct_id: mixpanelUUID,
+      ip,
+      ...param,
+    });
     await relaunchHeadless();
     event.returnValue = true;
   });
@@ -572,10 +577,11 @@ function initializeIpc() {
     event.returnValue = "Not supported platform.";
   });
 
-  ipcMain.on("mixpanel-track-event", async (_, eventName: string) => {
+  ipcMain.on("mixpanel-track-event", async (_, eventName: string, param: {}) => {
     mixpanel?.track(eventName, {
       distinct_id: mixpanelUUID,
       ip,
+      ...param,
     });
   });
 
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -544,7 +544,7 @@ function initializeIpc() {
     }
   });
 
-  ipcMain.on("relaunch standalone", async (event, param: {}) => {
+  ipcMain.on("relaunch standalone", async (event, param: object) => {
     mixpanel?.track("Launcher/Relaunch Headless", {
       distinct_id: mixpanelUUID,
       ip,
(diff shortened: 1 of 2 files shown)


src/main/main.ts
```diff
@@ -566,10 +571,11 @@ function initializeIpc() {
     event.returnValue = "Not supported platform.";
   });
 
-  ipcMain.on("mixpanel-track-event", async (_, eventName: string) => {
+  ipcMain.on("mixpanel-track-event", async (_, eventName: string, param: {}) => {
```
BasixKOR, line 574:
```suggestion
  ipcMain.on("mixpanel-track-event", async (_, eventName: string, param: object) => {
```

src/main/main.ts
```diff
@@ -538,7 +538,12 @@ function initializeIpc() {
     }
   });
 
-  ipcMain.on("relaunch standalone", async (event) => {
+  ipcMain.on("relaunch standalone", async (event, param: {}) => {
```
BasixKOR, line 541: The type `{}`, confusingly, includes primitive types such as numbers and strings.

Personally I think it's better to use dedicated type like `Record<"reason", string>`, but either way looks okay to me.

```suggestion
  ipcMain.on("relaunch standalone", async (event, param: object) => {
```

src/renderer/views/preload/PreloadProgressView.tsx
```diff
@@ -160,21 +160,21 @@ const PreloadProgressView = observer(() => {
...
           "Haven't received any messages for some time. Automatically relaunch."
         );
-        ipcRenderer.send("relaunch standalone");
+        ipcRenderer.send("relaunch standalone", {reason: "Haven't received message."});
         break;
       case 0x05:
         console.error("Action Timeout. Automatically relaunch.");
-        ipcRenderer.send("relaunch standalone");
+        ipcRenderer.send("relaunch standalone", {reason: "Action Timeout."});
         break;
```
BasixKOR, line 178: Seems like the fallback case is missing. Is it intended?

src/main/main.ts
riemannulus: I intend for give choice to caller. I guess `object` is looks good to me.

src/renderer/views/preload/PreloadProgressView.tsx
riemannulus: IDK why `default` fallback is missing. but I'm just do not fix what I don't know.

BasixKOR: Uh is it okay though? I think we might be losing some tracking information of some crashes with unknown errors.

riemannulus: I think that's fine. There is only one case of going to the default where the Launcher get an unexpected error code from Headless, but it doesn't look good for the launcher to handle an undefined error code arbitrarily.

BasixKOR: Well if that's the case I don't see a problem too. Thanks for an explaination!

src/main/main.ts
BasixKOR: Did you forget to replace this one?

BasixKOR (approved):

moreal (approved):

PR #788, opened 2021-06-29
Revise monster collection

longfin:
This PR adjusts GraphQL API to accept changes from https://github.com/planetarium/lib9c/pull/455.

- Due to `CancelMonsterCollect` unified into `MonsterCollect`, there is no `cancelMonsterCollect()` mutation no longer.
- `monsterCollectionState.end` is also removed, so we can check only level instead.
- `MonsterCollectionStatusType.canReceive` can be derived from `MonsterCollectionStatusType.rewardsInfo`. so it will be replaced too.

Although it is still in the middle of the work, I plan at this stage, we plan to organize and merge the PR once, then continue the follow-up work.

### DIFF
--- a/src/collection/pages/main/main.tsx
+++ b/src/collection/pages/main/main.tsx
@@ -8,7 +8,6 @@ import { Reward, CollectionItemTier, CollectionPhase, CollectionSheetItem } from
 import "./main.scss";
 import ConfirmationDialog from "../../components/ConfirmationDialog/ConfirmationDialog";
 import {
-  useCancelCollectionMutation,
   useCollectMutation,
   useGetTipQuery,
   useStagedTxQuery,
@@ -17,7 +16,6 @@ import {
   useCollectionStatusSubscription,
   useCollectionStateSubscription,
   useStateQueryMonsterCollectionQuery,
-  useCollectionStatusQueryQuery,
 } from "../../../generated/graphql";
 import { CollectionItemModel } from "../../models/collection";
 import ExpectedStatusBoard from "../../components/ExpectedStatusBoard/ExpectedStatusBoard";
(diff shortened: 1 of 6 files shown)


src/collection/pages/main/main.tsx
```diff
@@ -278,10 +273,10 @@ const Main: React.FC = () => {
       });
       return collectionResult.data!.action!.monsterCollect as string;
     } else if (data.stateQuery.agent!.monsterCollectionLevel > latestCollectionItem.tier) {
-      const collectionResult = await cancelCollection({
-        variables: { level: latestCollectionItem.tier },
+      const collectionResult = await collect({
+        variables: { level: 0 },
```
riemannulus, line 277: This logic is unnecessary unless `cancelCollect` and `collect` are separate. It seems better to fix the logic to do a `collect` only when it is not 0.

longfin: Amended. https://github.com/planetarium/9c-launcher/pull/788/commits/5e35dfa5afcc96ea53e9ef8181520c270a3c81e5

src/collection/pages/main/main.tsx
```diff
@@ -272,19 +267,11 @@ const Main: React.FC = () => {
...
-      const collectionResult = await cancelCollection({
-        variables: { level: latestCollectionItem.tier },
-      });
-      return collectionResult.data!.action!.cancelMonsterCollect as string;
-    } else {
-      return;
-    }
+    const collectionResult = await collect({
+      variables: { level: latestCollectionItem.tier },
+    });
```
riemannulus, line 272: IMO We need to check the level from GraphQL and `latestCollectionItem` is not the same. If levels are the same, returned null is the previous spec.

riemannulus (approved):

src/collection/pages/main/main.tsx
longfin: I'll add a check on the next PR.

PR #789, opened 2021-06-29
Improve webpack configuration

BasixKOR:
* Update webpack and used plugins and loaders to latest version
* Remove `@babel/polyfill` and replaced with `core-js`
* Update target to transpile to modern JS.
* Use Dart Sass instead of deprecated Node Sass.
* Builds seperate chunks for `node_modules` and `src/generated`.

This also enables proper source map, which was not available on previous configuration.

# Notiable changes
* It automatically cleans when build.
* **Assets such as images are now located on own folder instead of inlining it.**
* Due to use of `eval-cheap-module-source-map`, the build result is not human-readable, but now it builds faster.


BasixKOR:
# Possible improvements
* Move generated GraphQL codes to own chunks

riemannulus:
Did you test storybook too?

BasixKOR:
I forgot to test Storybook, but after some tweaking it looks like working well now!

riemannulus (approved):

PR #792, opened 2021-07-01
add offset argument to topmostblock query

area363:
This PR adds an `offset` argument for `topmostblocks` query so that the launcher will show relatively confirmed blocks mined instead of blocks that are highly likely to be reorged. The initial offset is set to 100(`confirmation: 2`) but I'm open to other values for this argument value.

Also, this should be merged after https://github.com/planetarium/NineChronicles.Headless/pull/542 is merged.

### DIFF
--- a/src/api.graphql
+++ b/src/api.graphql
@@ -24,7 +24,7 @@ query NodeStatus {
   }
 }
 
-query TopmostBlocks($limit: Int = 500, $offset: Int = 500, $miner: Address = null) {
+query TopmostBlocks($limit: Int = 500, $offset: Int = 100, $miner: Address = null) {
   nodeStatus {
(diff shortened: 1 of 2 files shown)


area363:
Could you take a look at this PR? It's related to https://github.com/planetarium/NineChronicles.Headless/pull/542 and it basically shows the number of blocks mined with an offset of 100 (`conf: 2`) on the launcher so that users are less confused about reorg :bow: @moreal @riemannulus @boscohyun

moreal (changes requested):
You should bump `NineChronicles.Headless` submodule to apply this feature.

area363:
Thanks for checking! I just updated to the latest development branch. PTAL @moreal :bow:

moreal (approved):

riemannulus (approved):

PR #795, opened 2021-07-02
fix(miningview): add missing text

BasixKOR:


src/renderer/views/login/MiningView.tsx
```diff
@@ -46,6 +46,14 @@ const MiningView = observer(
       }
     }, [requirementEl.current]);
 
+    const requirement: string = useT(
+      "REQUIRE:\n" +
```
longfin, line 50: Can we use `<br />` tag instead of `\n` even in near future?

longfin (approved):

src/renderer/views/login/MiningView.tsx
BasixKOR: I really wish we could do that. Unfortunately React wouldn't allow that, although we could tweak Transifex Native a little bit to make it happen.

PR #796, opened 2021-07-02
feat: new component to convert linebreak

BasixKOR:


src/renderer/views/main/MainView.tsx
```diff
@@ -26,10 +26,15 @@ const MainView = observer(
...
         <NineChroniclesLogo />
         <h1 className={classes.title}>
-          <T _str="Welcome to Nine Chronicles!" _tags={transifexTags}/>
+          <T _str="Welcome to Nine Chronicles!" _tags={transifexTags} />
         </h1>
         <article className={classes.body}>
-          <T _str="This is a fantasy world on the blockchain.\nYou need to create an account to start the game." _tags={transifexTags}/>
+          <T
+            _str={
+              "This is a fantasy world on the blockchain.\nYou need to create an account to start the game."
```
longfin, line 34:
```suggestion
              "This is a fantasy world on the blockchain.\n" + 
              "You need to create an account to start the game."
```

src/renderer/i18n/index.tsx
```diff
@@ -44,3 +50,32 @@ export async function validateLocale(locale: string): Promise<boolean> {
...
+
+export function T<Content = {}>({
+  _str,
+  ...props
+}: TransifexProps<Content>): JSX.Element {
+  const node: ReactChild = useT(_str, props);
+  return <>{newlineToBreak(node)}</>;
+}
+
+function newlineToBreak(str: ReactChild): ReactNode {
```
moreal, line 67: IMO it should use *line break* instead of *break*.

```suggestion
function newlineToLineBreak(str: ReactChild): ReactNode {
```

moreal: IMO the parameter name, `str` should be replaced with `node` or `child`. Its name `str` feels like a string value but the `ReactChild` type is same with `string | number | ReactElement` and it cannot assume the `ReactChild` will be always `string` type.

BasixKOR: I agree with you. I guess I forgot to rename it while working on it...

riemannulus (approved):

riemannulus:
It seems need rebase 😢

longfin (approved):

PR #807, opened 2021-07-08
Consider `pt_BR`

boscohyun:
SSIA

### DIFF
--- a/src/renderer/i18n/index.tsx
+++ b/src/renderer/i18n/index.tsx
@@ -40,7 +40,13 @@ export function LocaleProvider({ children }: React.PropsWithChildren<{}>) {
 
 export async function validateLocale(locale: string): Promise<boolean> {
   const languages = await tx.getLanguages();
-  return languages.some((lang: Record<"code", string>) =>
-    locale.startsWith(lang.code)
-  );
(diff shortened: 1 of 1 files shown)


src/renderer/i18n/index.tsx
```diff
@@ -40,7 +40,13 @@ export function LocaleProvider({ children }: React.PropsWithChildren<{}>) {
...
-    locale.startsWith(lang.code)
-  );
+  return languages.some((lang: Record<"code", string>) => {
+    let code = lang.code;
+    if (code === "pt_BR")
+    {
+      code = "pt-BR";
+    }
+
+    return locale.startsWith(code);
```
riemannulus, line 50: IMO this problem will happen with other `Unix style tagging`. How about change this code like...
```suggestion
    return locale.startsWith(lang.code.replace('_', '-'));
```

riemannulus (commented):
LGTM but

src/renderer/i18n/index.tsx
boscohyun: That's batter.😁

riemannulus (approved):
👍

moreal (approved):

PR #811, opened 2021-07-08
Fix the gold sync

longfin:
This PR fixes the gold synchronization bug in monster collection related UI. To accomplish this, I removed `GoldAndCollectionLevel` and unified `gold` field to `collectionStateQuery` mainly polled.

### DIFF
--- a/src/renderer/components/AccountInfo/AccountInfoContainer.tsx
+++ b/src/renderer/components/AccountInfo/AccountInfoContainer.tsx
@@ -36,6 +36,7 @@ const AccountInfoContainer: React.FC<Props> = (props: Props) => {
   const [isCollecting, setIsCollecting] = useState<boolean>(false);
   const [canClaim, setCanClaim] = useState<boolean>(false);
   const [collectionLevel, setCollectionLevel] = useState<number>(0);
+  const [receivedBlockIndex, setReceivedBlockIndex] = useState<number>(0);
   const {
     refetch: sheetRefetch,
   } = useCollectionSheetQuery();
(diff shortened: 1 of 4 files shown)


boscohyun (approved):

src/renderer/components/AccountInfo/AccountInfoContainer.tsx
```diff
@@ -178,11 +165,7 @@ const AccountInfoContainer: React.FC<Props> = (props: Props) => {
               ? () => { }
               : onOpenWindow}
           canClaimReward={mcStatus?.rewardInfos != null && mcStatus.rewardInfos.length > 0}
-          goldLabel={
-            mcStatus?.fungibleAssetValue.quantity
-              ? Number(mcStatus.fungibleAssetValue.quantity)
-              : Number(goldAndLevel?.stateQuery.agent?.gold)
-          }
+          goldLabel={Number(collectionStateQuery?.stateQuery.agent?.gold)}
```
riemannulus, line 168: Does this change the gold? I don't seem to see any code to refetch the Query.

longfin: [`collectionStateQuery` will be polled.](https://github.com/planetarium/9c-launcher/blob/a103dfa457da25e7f0c6fae1e213c1788a1a4e00/src/renderer/components/AccountInfo/AccountInfoContainer.tsx#L53) so it will be changed.

riemannulus (approved):

src/renderer/components/AccountInfo/AccountInfoContainer.tsx
riemannulus: So, now we do not use `subscription` anymore?

longfin: subscription is still used when we receive messages between polling cycles.

longfin:
I've added two commits to address the loading state sync problem. PTAL @boscohyun @riemannulus

boscohyun (approved):

PR #813, opened 2021-07-09
feat(main): add cause params to HeadlessInitializeError

BasixKOR:
Also formatted entire main.ts, I couldn't find a way to format specific range only.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -59,7 +59,7 @@ import { DownloadSnapshotFailedError } from "./exceptions/download-snapshot-fail
 import { DownloadSnapshotMetadataFailedError } from "./exceptions/download-snapshot-metadata-failed";
 import { ClearCacheException } from "./exceptions/clear-cache-exception";
 import createCollectionWindow from "../collection/window";
-import { Client as NTPClient } from 'ntp-time'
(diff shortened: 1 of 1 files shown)


src/errors/HeadlessInitializeError.ts
```diff
@@ -1,6 +1,15 @@
...
-    super(msg);
+  public cause: Error | null;
+
+  constructor(msg: string, cause?: Error) {
+    const message = `${msg} ${cause ? `caused by ${cause.message}` : ''}`;
+    super(message);
+
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, HeadlessInitializeError);
+    }
```
BasixKOR, line 10: See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#custom_error_types

riemannulus (approved):

riemannulus:
Rebase please @BasixKOR

riemannulus (approved):

BasixKOR:
Note: When tc39/proposal-error-cause gets to Stage 4, we could refactor this to use the standard cause property instead.

longfin (approved):

PR #820, opened 2021-07-14
Enables remote config

longfin:
This PR enables config loads from a remote location (`https://download.nine-chronicles.com/9c-launcher-config.json`). furthermore, it adds `ConfigVersion` to opt-out of this feature by a user.

And, I renamed symbols related to the launcher configuration. (it's too confusing 😅 ) too.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -102,6 +102,10 @@ export const electronStore = new Store<IElectronStore>({
     DataProviderUrl: {
       type: "string",
       default: undefined
+    },
+    ConfigVersion: {
+      type: "number",
+      default: 0
     }
   },
 });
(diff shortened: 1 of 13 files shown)


ipdae (approved):

jckdotim (approved):
LGTM

BasixKOR:
Could we leverage `migrations` feature to keep user-edited contents?

https://github.com/sindresorhus/electron-store#migrations

longfin:
> Could we leverage `migrations` feature to keep user-edited contents?
> 
> https://github.com/sindresorhus/electron-store#migrations

migration on electron-store seems to be [schema migration](https://en.wikipedia.org/wiki/Schema_migration). Since `ConfigVersion` is just for preserving the user's previous settings, I'm not sure if that's the right thing to use. 😅  (It is also scheduled to be reset when APV rises)

Issue #579, opened 2021-03-12
Config should not be initialized after every update


limebell:
Updater now overwrites `config.json` file every update, so user custom configuration is initialized. Especially headless node is now start up automatically so blockchain is stored unintended place and its size is very big.

PR #823, opened 2021-07-15
Separate settings into app settings and user settings

longfin:
This PR is the following work of #820 .

The problem with #820 is that you can't persist the local settings without a `ConfigVersion`. The 9c launcher has a UI to change settings and is an app that users frequently change settings for, and if you change the ConfigVersion each time, the goal of #820 is practically difficult to achieve.

Therefore, to fundamentally solve this problem,
- Instead of following the existing app installation path, `config.json`, create a separate `config.json` left in the user setting directory.
- Arrange the code so that all settings can only be accessed through `get()` defined in `config.ts`
- Modified entire code to lookup user settings first in `get()`, and follow app settings only if not present.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -5,110 +5,113 @@ import { IConfig } from "./interfaces/config";
 const { app } =
   process.type === "browser" ? require("electron") : require("electron").remote;
 
+const schema: any = {
+  AppProtocolVersion: {
+    type: "string",
+    default:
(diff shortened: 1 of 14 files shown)


riemannulus (approved):

BasixKOR (approved):

BasixKOR:
Closes #579

BasixKOR:
Related: #579

PR #828, opened 2021-07-19
Close stream when canceled with cancellationToken

riemannulus:
We assume that the current download failure-related symptom is that the stream is not closed properly and the download continues, so we take an experimental measure.


BasixKOR (approved):

BasixKOR (changes requested):
Turns out the Electron version we're using ships with Node v12, which doesn't have Promise interface (they were introduced in Node v15).

We should consider other mesaures such as destroying pipe when cancelled, which was a previously considered solution.

PR #834, opened 2021-07-27
Bump headless

limebell:
This bump includes behavior option when the chain tip is stale, and bucket size configuration.


longfin:
Is it okay we don't bump lib9c and libplanet?

limebell:
> Is it okay we don't bump lib9c and libplanet?

It is most likely okay but I'm going to bump it because the workflow fails.

longfin (approved):

PR #839, opened 2021-07-28
Initialize v2

BasixKOR:
This PR includes initial cornerstone for v2 development, primarily the basic layout.

# Notable changes
* Design refresh
* Use Apollo Client v3
  * Custom link for more reliable connection with Headless (Implementation only, TBD)
* Uses XState-based FSM in appropriate places
* Uses CSS Modules
* Avoids Material UI
  
# Reused from v1
* MobX stores
* i18n implementation

# Note
Please run this if you're Windows user, since Git For Windows disables symlink support by default: `git config core.symlinks true`


BasixKOR:
image

Currently it looks something like this. It handles popup-ish UI as well but the details aren't implemented here now.

longfin:
`src/v2/generated/graphql.tsx` doesn't seem to need version control.

PR #840, opened 2021-07-29
Fix devtools to open correctly

BasixKOR:


riemannulus:
Could you refactor this code too? https://github.com/planetarium/9c-launcher/blob/development/src/collection/window/index.ts

riemannulus (approved):

riemannulus:
need rebase too

longfin (approved):

PR #875, opened 2021-08-13
Tracking login status on mixpanel

riemannulus:
Add a Mixpanel event to check if the user exits after logging in, or before logging in.


src/main/main.ts
```diff
@@ -106,6 +103,12 @@ export type MixpanelInfo = {
 
 ipv4().then((value) => (ip = value));
 
+const mixpanelUUID = loadInstallerMixpanelUUID();
+const mixpanel: NineChroniclesMixpanel | undefined =
+  getConfig("Mixpanel") && !isDev
+    ? new NineChroniclesMixpanel(createMixpanel(MIXPANEL_TOKEN), mixpanelUUID, ip || "unknown")
```
BasixKOR, line 103: When we supplied `ip` for every track call it had no problem because eventually the value will be filled. However in this case the IP value may not be initialized by the time the NineChroniclesMixpanel instance is created (and I think it is likely to happen), and will keep uninitialized value for rest of runtime.

I'd suggest calling `ipv4` after intializing the instance, and call the setter to update the IP address.

src/main/mixpanel.ts
```diff
@@ -0,0 +1,45 @@
+import { Mixpanel } from "mixpanel";
+
+export interface INineChroniclesMixpanel {
```
BasixKOR, line 3: Do we have a reason to declare interface rather than using class type directly?

BasixKOR (commented):
Using wrapper for Mixpanel looks nice to me, but I got some suggestions :)

src/main/main.ts
riemannulus: I realized `ip` is using only here. how about `ip` initialize on the `NineChroniclesMixpanel`?

BasixKOR: Yeah, that looks better for me.

src/main/mixpanel.ts
riemannulus: This code is intended to make writing test code a little easier. I thought this would make it easier to isolate dependencies.

BasixKOR: You don't need to worry about that case, since TypeScript uses structual typing on classes just like interfaces.

https://www.typescriptlang.org/docs/handbook/2/classes.html#relationships-between-classes

riemannulus: That is awesome! But in this case, I think it is easy to get help from the IDE which Mocking class has the same interface as the corresponding class.

riemannulus: You can check [__tests__/mock/MockNineChroniclesMixpanel.ts](https://github.com/planetarium/9c-launcher/pull/875/commits/4cad1598e42db4ab6f9df14a04b5bfef23e07d13) too.

src/main/main.ts
riemannulus: 5a04499 amended.

riemannulus:
Test passed. PTAL 🙏 @BasixKOR

src/main/mixpanel.ts
```diff
@@ -0,0 +1,48 @@
...
+    private readonly _ip: string;
+    private _login: boolean = false;
+    private _miningConfig: boolean = false;
+
+    constructor(mixpanel: Mixpanel, mixpanelUuid: string) {
+        this._mixpanel = mixpanel;
+        this._mixpanelUuid = mixpanelUuid;
+        let ip: string = 'unknown'
+        ipv4().then((value) => (ip = value));
+        this._ip = ip;
```
BasixKOR, line 23:
```suggestion
        ipv4().then((value) => (this._ip = value));
```

.then() is called as a microtask, which means the handler will run **after** the constructor call. This will result in `_ip` being always unknown.

BasixKOR (changes requested):
Although moving IP fetching inside the class was nice, there are some changes to be made to ensure correct behaviour. Could you take a look into those as well?

src/main/mixpanel.ts
riemannulus: But `this._ip` is `readonly` property. so we cannot init here. Have you a good idea about this?

BasixKOR: For now just keep it mutable would be reasonable enough. It was already managed as mutable value before, so I think it is a reasonable compromise.

riemannulus: Hmm... ok, I'll make this variable as a mutable. But I guess we need some comments here like 'This variable must not be changed.` to prevent mistakes in other engineers.

riemannulus: apply on f074d11

BasixKOR (approved):

src/main/mixpanel.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    public login(): void {
+        this._login = true;
+    }
+
+    public miningConfig(): void {
+        this._miningConfig = true;
+    }
+
+}
```
longfin, line 48:
```suggestion
}

```

src/renderer/views/login/MiningView.tsx
```diff
@@ -23,6 +23,7 @@ const MiningView = observer(
       if (setSucceed) {
         standaloneStore.setPrivateKeyEnded(true);
         accountStore.setMiningConfigStatus(true);
+        ipcRenderer.send("set mining");
```
longfin, line 26: Is this change intended? it doesn't seem to be related to MixPanel.

riemannulus: It is intended. see also. 
https://github.com/planetarium/9c-launcher/pull/875/files#diff-7adaa52a2eacb085ba5e07e6683a61fb3601547dc00b51fd50fbaf685e13fe96R544-R547

src/main/mixpanel.ts
riemannulus: fixed. 90068fa

src/renderer/views/login/MiningView.tsx
longfin: 🆗

longfin (approved):

Issue #851, opened 2021-08-04
(POC) use remote headless


ipdae:
- Configure rpc config in config.json
- Use remote(local) headless instead of starting headless

PR #881, opened 2021-08-17
Introduce rpc config

ipdae:
- introduce remote rpc mode by local headless. #851

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -112,6 +112,26 @@ const schema: any = {
   ConfigVersion: {
     type: "integer",
     default: 0
+  },
+  RemoteRpcServerHost: {
+    type: "string",
+    default: "127.0.0.1"
+  },
+  RemoteRpcServerPort: {
+    type: "integer",
+    default: 23042
+  },
+  RemoteGraphQLServerHost: {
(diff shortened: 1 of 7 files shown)


area363:
There's an E2E Test error 😢 (https://github.com/planetarium/9c-launcher/pull/881/checks?check_run_id=3349208033#step:9:11)

ipdae:
@area363 fix error.

area363 (approved):

src/config.ts
```diff
@@ -129,6 +149,14 @@ const GraphQLServer = (): string => {
   return `${LocalServerUrl}/graphql`;
 };
 
+const RemoteGraphQLServer = (): string => {
+  return `${RemoteHeadlessUrl}/graphql`;
+}
+
+const RemoteHeadlessUrl = (): string => {
+  return `${get("RemoteGraphQLServerHost")}:${get("RemoteGraphQLServerPort")}`;
```
BasixKOR, line 157: Could we use somthing like `RemoteGraphQLServerURL` instead of specifiying parts of it? It seems like it's always used as URL form.

PR #908, opened 2021-09-01
feat(transfer): add notice on swap to wncg page

riemannulus:
image


Namyujeong (approved):
I fixed the notice font color

PR #919, opened 2021-09-08
feat(transfer): update swap maximum per daily

moreal:
Since yesterday, NineChronicles Ethereum Bridge's swap maximum limit was down to 5,000 NCG per day.
This pull request applies the change to 9c-launcher.

### DIFF
--- a/src/transfer/pages/swap/swap.tsx
+++ b/src/transfer/pages/swap/swap.tsx
@@ -148,7 +148,7 @@ const SwapPage: React.FC<Props> = observer((props: Props) => {
           </li>
           <li>
             <SwapNoticeLabel>
-              <T _str="* Maximum 5,000 NCG per day" _tags={transifexTags} />
+              <T _str="* Maximum {max, number, integer} NCG per day" _tags={transifexTags} max={5000} />
             </SwapNoticeLabel>
           </li>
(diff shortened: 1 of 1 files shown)


BasixKOR (approved):
I think we gotta change the value from hardcoded to be interpolated sometime in a future, but for now this seems nice enough.

src/transfer/pages/swap/swap.tsx
```diff
@@ -148,7 +148,7 @@ const SwapPage: React.FC<Props> = observer((props: Props) => {
           </li>
           <li>
             <SwapNoticeLabel>
-              <T _str="* Maximum 100,000 NCG per day" _tags={transifexTags} />
+              <T _str="* Maximum 5,000 NCG per day" _tags={transifexTags} />
```
BasixKOR, line 151:
```suggestion
              <T _str="* Maximum {max, number, integer} NCG per day" _tags={transifexTags} max={5000} />
```

This code will show the same message, but allows to change maximum value without causing to retranslate same pharse.

moreal: I applied it in 203fb86. 🙏🏻

BasixKOR (approved):
LGTM

PR #923, opened 2021-09-10
chore: bump transifex libs

BasixKOR:
Transifex will be sunsetting support for old Transifex Native SDKs starting in early September.


longfin (approved):

longfin:
it has conflicts.

PR #930, opened 2021-09-13
V2 (New Interface)

BasixKOR:
It was meant to be a tempoary branch for holding future developments to be broken down to own branches later on, but as the v2 initialization branch's merge got delayed there was too much content on this branch to be inspected.

This branch currently holds:
 * Literally everything V2


BasixKOR:
Blocked-by: #933

BasixKOR:
# Application-wide changes
Due to some technical implications, some changes have been made across the application. Here is the list of the changes that have been made to make V2 work well with the current launcher code base.

* Some libraries are bumped.
  * `subscriptions-transport-ws` has been bumped. This may have an impact on GraphQL subscriptions, although I don't expect any major issues.
  * `@babel/core` has been bumped to parse more recent TypeScript types. This won't have much impact as long as no bugs are introduced to the newer version.
  * **`react`, `react-dom` has been bumped to 17.x.** As the React team describes this should have minimal impact but due to the nature of the changes it makes, it should be verified.
* `UseV2Interface` configuration item has been added. This will be used to check if the user has opted-in to the newer V2 interface until the migration completes.
* Apollo Devtools has been added on development builds to help identify GraphQL issues.
* `generate-private-key` IPC call has been added to make onboarding UX better by providing the user ID while creating a new ID.
* Some V2-specific IPC events are added and it will be used for V2-specific needs. It will be ignored on the current launcher.
* Some changes are made to the older launcher code base but it is mostly stylistic changes that are made because of Prettier version differences.
* Some TypeScript settings have been adjusted to make Storybook work with V2 interfaces. It should have minimal impact on existing stories since we aren't using Emotion.

Namyujeong (approved):

PR #945, opened 2021-09-17
Fix CI to check coding styles

BasixKOR:


longfin:
- There is a conflict. 😢 
- Could you explain why it's okay to remove `pretty-quick`?

BasixKOR:
* Unfortunely `yarn.lock` conflicts aren't easy to fix (and its auto-generated) so I'm planning to merge it locally and resolve conflicts by `yarn`, which automatically resolves conflict.
* The prettier itself has capablities to format and check code styles, and `pretty-quick` for some reason doesn't check changed codes in CI and just passes. Using Prettier would result in slightly longer CI time though.

BasixKOR:
It seems like CI and my local environment disagrees with each other. I'll elevate `--loglevel` for debug for debugging.

PR #955, opened 2021-09-27
improve confirmation when clicking clear cache

BasixKOR:
image

### DIFF
--- a/src/renderer/components/ClearCacheButton.tsx
+++ b/src/renderer/components/ClearCacheButton.tsx
@@ -6,7 +6,7 @@ import DeleteIcon from "@material-ui/icons/Delete";
 import CheckCircleIcon from "@material-ui/icons/CheckCircle";
 
 const { dialog } = remote;
-const message = t("All block data will be deleted and re-downloaded again. This process can take up to 30 minutes. Are you sure to continue?");
+const message = t("All local chain data will be deleted and re-downloaded again. This process can take up to 30 minutes. Are you sure to continue?");
 
 interface IClearCacheButtonProps extends ButtonProps {
   disabled?: boolean;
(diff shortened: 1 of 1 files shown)


src/renderer/components/ClearCacheButton.tsx
```diff
@@ -1,18 +1,29 @@
 import React, { useState } from "react";
-import { ipcRenderer } from "electron";
+import { ipcRenderer, remote } from "electron";
 import { Button, ButtonProps } from "@material-ui/core";
+import { t } from "@transifex/native";
 import DeleteIcon from "@material-ui/icons/Delete";
 import CheckCircleIcon from "@material-ui/icons/CheckCircle";
 
+const { dialog } = remote;
+const message = t("All block data will be deleted and re-downloaded again. This process can take up to 30 minutes. Are you sure to continue?");
```
longfin, line 9:
```suggestion
const message = t("All local chain data will be deleted and re-downloaded again. This process can take up to 30 minutes. Are you sure to continue?");
```

src/renderer/components/ClearCacheButton.tsx
```diff
@@ -1,18 +1,29 @@
...
 
 const ClearCacheButton = (props: IClearCacheButtonProps) => {
   const [isCleared, setClearState] = useState(false);
 
-  const handleClick = () => {
-    if (window.confirm("Are you sure you want to clear cache?")) {
+  const handleClick = async () => {
+    const result = await dialog.showMessageBox({
+      type: "question",
+      buttons: ["Yes", "No"],
```
longfin, line 21: It would be better if we also translate these button labels too.

longfin (approved):

src/renderer/components/ClearCacheButton.tsx
BasixKOR: It seems like the Windows (or Electron, unsure) automatically translates common button labels like this one. It can be seen in the screenshot above.

Namyujeong (approved):

PR #962, opened 2021-09-30
Support monster collection in rpc mode

ipdae:
- support monster collection in rpc mode
- resolve https://github.com/planetarium/NineChronicles.Headless/issues/709


src/renderer/components/ClaimCollectionRewardDialog/ClaimCollectionRewardDialog.tsx
```diff
@@ -32,6 +53,49 @@ const ClaimCollectionRewardDialog: React.FC<Props> = (props: Props) => {
...
+    // create action.
+    const fileName = await tmpName();
+    if (avatarAddress.startsWith("0x"))
+    {
+      avatarAddress = avatarAddress.substr(2);
+    }
+    if (!ipcRenderer.sendSync("claim-monster-collection-reward", avatarAddress, fileName))
+    {
+      console.error(`makeTx: failed action`)
+      return "";
```
longfin, line 68: How about throwing an error instead of returning an empty string?

PR #1078, opened 2021-11-30
filter node list by node status

### LINKED ISSUE #1072: Filter preload ended nodes only in node list

ipdae:
- resolve #1072 
- Fix `NodeInfo` select. using ipc communication instead of config variables.
- refactoring app initialize for waiting node list.

before
```
13:28:48.231 › Executing the remote headless server:
  a0d01f897e0a2434798e8b4607ac32ea-994288696.us-east-2.elb.amazonaws.com/31238
1:30
Request URL: http://adb1932b4da92426abd7116c65875faa-1809698636.us-east-2.elb.amazonaws.com/graphql
```

after
```
18:47:58.718 › Executing the remote headless server:
  aba105ce4868747839ab806175be8b37-13503424.us-east-2.elb.amazonaws.com/31238
Request URL: http://aba105ce4868747839ab806175be8b37-13503424.us-east-2.elb.amazonaws.com/graphql

```

### DIFF
--- a/src/collection/App.tsx
+++ b/src/collection/App.tsx
@@ -4,7 +4,7 @@ import { split, ApolloLink } from "apollo-link";
 import { createHttpLink } from "apollo-link-http";
 import { RetryLink } from "apollo-link-retry";
 import { WebSocketLink } from "apollo-link-ws";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import "./App.scss";
 import { getMainDefinition } from "apollo-utilities";
 import { ApolloProvider } from "react-apollo";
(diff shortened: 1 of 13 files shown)


src/renderer/App.tsx
```diff
@@ -117,6 +75,59 @@ function App() {
...
+    wsLink,
+    httpLink
+  );
+
+  const link = ApolloLink.from([new RetryLink(), apiLink]);
+
+  const client = new ApolloClient({
+    link: link,
+    cache: new InMemoryCache(),
+  });
```
BasixKOR, line 115: These changes will cause a creation of a new Apollo Client on every render, which is highly undesirable. I know we gotta get URL via IPC first but still we gotta find a better way.

src/collection/App.tsx
```diff
@@ -63,19 +39,49 @@ function createFile() {
...
+        definition.operation === "subscription"
+      );
+    },
+    wsLink,
+    httpLink
+  );
+
+  const link = ApolloLink.from([new RetryLink(), apiLink]);
+
+  const client = new ApolloClient({
```
BasixKOR, line 80: See comments on `src/renderer/App.tsx`.

src/main/main.ts
```diff
@@ -551,9 +567,10 @@ function initializeIpc() {
     utils.deleteBlockchainStoreSync(getBlockChainStorePath());
     if (rerun) {
       if (useRemoteHeadless) {
-        initializeRemoteHeadless();
+        console.log("main clear cache call initializeRemoteHeadless");
+        await initializeRemoteHeadless();
       } else {
-        initializeHeadless();
+        await initializeHeadless();
```
BasixKOR, line 573: Just a curiosity, what does this do?

src/renderer/stores/game.ts
```diff
@@ -43,26 +43,18 @@ export default class GameStore {
   }
 
   @action
-  startGame = (privateKey: string) => {
+  startGame = async (privateKey: string) => {
```
BasixKOR, line 46: I couldn't find a call to `startGame` using Promise or `await`. Is this change intended?

BasixKOR (changes requested):
Thank you for your hard work. However, there are some changes that need further discussion IMO.

src/v2/utils/useStore.tsx
```diff
@@ -6,7 +6,6 @@ import StandaloneStore from "../stores/standaloneStore";
 
 const stores = {
   account: new AccountStore(),
-  game: new GameStore(),
```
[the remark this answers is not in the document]
ipdae: I found it when change the GameStore constructor, there was no place currently used the stores.GameStore.

src/renderer/stores/game.ts
ipdae: It seems previous work remain. I'll delete it.

src/main/main.ts
ipdae: It seems previous work remain. I'll delete it.

src/renderer/stores/game.ts
ipdae: It seems previous work remain. I'll delete it.

src/main/main.ts
```diff
@@ -103,6 +105,7 @@ const client = new NTPClient("time.google.com", 123, { timeout: 5000 });
 
 let remoteHeadless: RemoteHeadless;
 let useRemoteHeadless: boolean;
+let remoteNode: NodeInfo;
```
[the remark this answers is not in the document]
ipdae: How can I move Apollo initialization?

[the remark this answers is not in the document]
BasixKOR: Assuming we have `remoteNode` determined at the start of process and it doesn't change afterwards, we could make a quick handler like this in main process:

```ts
ipcMain.on("get remote node", event => event.returnValue = remoteNode);
```

and call them as sendSync like this: `ipcRenderer.sendSync("get remote node")`. Of course this has performance implications as well but since we are just returning same value and we wouldn't make a lot of calls to them I think it should be fine.

src/config.ts
```diff
@@ -197,35 +187,58 @@ export class NodeInfo {
...
+      `http://${this.GraphqlServer()}`,
+      {
+        query: "query{nodeStatus{preloadEnded}}",
+      },
+      {
+        headers: {
+          "Content-Type": "application/json",
+        },
+      }
+    );
```
BasixKOR, line 202: Just a note but we can reuse `graphql-request` module here since we already have it as a dependency.

src/collection/App.tsx
```diff
@@ -41,47 +41,48 @@ const isFileExsist = getIsFileExsist();
 
 const App: React.FC = () => {
   const [agentAddress, setAgentAddress] = useState<string>("");
-  const [headlessUrl, setHeadlessUrl] = useState<string>("");
+  const [client, setClient] = useState<ApolloClient<any>>();
 
-  ipcRenderer.on("initialize collection window", (_, address, node) => {
+  ipcRenderer.on("initialize collection window", (_, address, headlessUrl) => {
```
BasixKOR, line 46: Could we wrap this code in useEffect? I'll do that if you don't mind.

BasixKOR (approved):
Overall nice.

PR #1093, opened 2021-12-07
Initialize gold balance before render block

### LINKED ISSUE #1091: invalid gold label when initialize
image

ipdae:
- Fix local remote node work
- resolve #1091

### DIFF
--- a/src/renderer/components/AccountInfo/AccountInfoContainer.tsx
+++ b/src/renderer/components/AccountInfo/AccountInfoContainer.tsx
@@ -41,6 +41,7 @@ const AccountInfoContainer: React.FC<Props> = (props: Props) => {
   const [collectionLevel, setCollectionLevel] = useState<number>(0);
   const [receivedBlockIndex, setReceivedBlockIndex] = useState<number>(0);
   const [tip, setTip] = useState<number>(0);
+  const [gold, setGold] = useState<number>(0);
   const [
     avatarAddressQueryLazy,
     { loading: avatarLoading, data: avatars, refetch: avatarRefetch },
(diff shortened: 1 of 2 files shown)


BasixKOR (approved):

src/renderer/components/AccountInfo/AccountInfoContainer.tsx
```diff
@@ -84,6 +85,13 @@ const AccountInfoContainer: React.FC<Props> = (props: Props) => {
       currentReward.set(x!.itemId, x!.quantity);
     });
     setCurrentReward(currentReward);
+
+    const gold =
+      collectionStatus?.monsterCollectionStatusByAgent.fungibleAssetValue
+        .quantity ??
+      collectionStatusQuery?.monsterCollectionStatus?.fungibleAssetValue
+        .quantity;
+    setGold(Number(gold));
```
BasixKOR, line 94: This could be solved by useMemo too:

```typescript
const gold = useMemo(() =>
      collectionStatus?.monsterCollectionStatusByAgent.fungibleAssetValue
        .quantity ??
      collectionStatusQuery?.monsterCollectionStatus?.fungibleAssetValue
        .quantity, [collectionStatus, collectionStatusQuery]);

ipdae: i'll try this.

BasixKOR (approved):
Nice! 😊

BasixKOR:
I think this is the cause of broken test: https://github.com/planetarium/9c-launcher/blob/7c75ffc81f3de99bef9af09babeb7b6ef78e3bd3/src/renderer/components/AccountInfo/AccountInfo.tsx#L30

It seems like it is failing now because we have no initial value (which was zero before). Maybe appending `?? 0` could work?

PR #1115, opened 2021-12-15
Handle when all remote node are unavailalble.

ipdae:
image

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -330,6 +330,9 @@ export const TRANSIFEX_TOKEN = "1/9ac6d0a1efcda679e72e470221e71f4b0497f7ab";
 export async function initializeNode(): Promise<NodeInfo> {
   console.log("config initialize called");
   const nodeList = await NodeList();
+  if (nodeList.length < 1) {
+    throw Error("can't find available remote node.");
+  }
   nodeList.sort((a, b) => {
     return a.clientCount - b.clientCount;
   });
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -205,7 +205,17 @@ async function initializeApp() {
     if (app.commandLine.hasSwitch("v2")) win = await createV2Window();
     else win = await createWindow();
     createTray(path.join(app.getAppPath(), logoImage));
-    remoteNode = await initializeNode();
+    try {
+      remoteNode = await initializeNode();
+    } catch (e) {
+      await dialog.showMessageBox(win!, {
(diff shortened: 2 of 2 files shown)


src/config.ts
```diff
@@ -330,6 +331,9 @@ export const TRANSIFEX_TOKEN = "1/9ac6d0a1efcda679e72e470221e71f4b0497f7ab";
 export async function initializeNode(): Promise<NodeInfo> {
   console.log("config initialize called");
   const nodeList = await NodeList();
+  if (nodeList.length < 1) {
+    throw new InitializeNodeFailedError();
+  }
   nodeList.sort((a, b) => {
     return a.clientCount - b.clientCount;
   });
```
BasixKOR, line 339: Not really related to this PR per se, but do we need to sort this list if we're going to pick randomly?

ipdae: yes, this is temporary loadbalance in client side before rpc node support alb(or nlb).

src/main/main.ts
```diff
@@ -205,7 +205,17 @@ async function initializeApp() {
...
-    remoteNode = await initializeNode();
+    try {
+      remoteNode = await initializeNode();
+    } catch (e) {
+      await dialog.showMessageBox(win!, {
+        message: "Failed to connect remote node. please restart launcher.",
+        type: "error",
+      });
+
+      win.webContents.send("go to error page", "initialize-node-failed");
```
BasixKOR, line 216: This line doesn't really do anything since it will be exited immediately.

src/main/main.ts
```diff
@@ -205,7 +205,17 @@ async function initializeApp() {
     if (app.commandLine.hasSwitch("v2")) win = await createV2Window();
     else win = await createWindow();
     createTray(path.join(app.getAppPath(), logoImage));
-    remoteNode = await initializeNode();
+    try {
+      remoteNode = await initializeNode();
+    } catch (e) {
```
longfin, line 210: How about adding a log related to `e`? (i.e. `console.error(e)`)

longfin (approved):

src/main/main.ts
BasixKOR: Nice catch. It would be better if the thrown error on `config.ts` included more information about the error.

moreal (approved):

area363 (approved):
👍

longfin (approved):

PR #1128, opened 2021-12-21
Add spinner while loading RPC nodes

BasixKOR:
image


BasixKOR:
Well, there's a spinner in there; it's just screenshot was taken at such a weird timing

ipdae (approved):

PR #1162, opened 2022-01-14
Move updates into own file

### LINKED ISSUE #1159: Check updates before connecting to a RPC node
## Problem
The launcher tries to connect to an available RPC node first before checking the APV. This makes the launcher unable to update when the RPC nodes are unavailable even though it is technically possible.

## Solution
Move the version checking out of `initializeHeadless`.

BasixKOR:
Also fixes #1159.


ipdae (approved):

src/main/main.ts
```diff
@@ -154,6 +150,7 @@ if (!app.requestSingleInstanceLock()) {
 
   cleanUp();
 
+  checkForUpdates(standalone);
```
BasixKOR, line 153: We should pass the `Update` returned from `checkForUpdates` to the `update` function. The result is unused here.

PR #1170, opened 2022-01-19
bump submodules

area363:
SSIA


BasixKOR:
Although I have no idea why the Circle CI checks are gone for this particular PR (I'll let them know) but the commit hashes of Lib9c submodules should be the same.

area363:
@BasixKOR I'll push to upstream from now on 😅 I checked the lib9c commits and they are the same so I'll merge this PR cc @ipdae

ipdae (approved):
check lib9c hash is same.

PR #1173, opened 2022-01-21
Initialize Refiner

BasixKOR:
In this PR we introduce a new user survey tool for getting better feedback from our users, directly. We expect this PR to enable us to have better communication with our users and have better insight into what our users are expecting from the launcher and Nine Chronicles as a whole.


moreal (commented):
It seems it has a style issue (from prettier)

moreal (approved):

area363 (approved):

PR #1177, opened 2022-01-22
Add description about git submodle download and skip build unity player

x86chi:


BasixKOR (approved):
Thank you for improving the documentation! Adding `--recursive` to the code snippet in the 'Setting things up' chapter would be also appreciated.

PR #1184, opened 2022-01-28
Show RPC hostname on info pane

### LINKED ISSUE #1167: Show RPC server info on the right pane
image

it would be better if we have detailed info about connected RPC server.

BasixKOR:
Fixes #1167.


area363:
@BasixKOR Are you bumping submodules as well?

BasixKOR:
@area363 No it was pure mistake. I'm correcting them.

area363 (approved):

PR #1246, opened 2022-03-10
Various V2 Monster Collection fixes

BasixKOR:
- Not show the overlay when it isn't needed.
 - "Get Rewards" button no longer triggers the monster collection window.
 - Loading state is now much clear to identify.

### DIFF
--- a/src/v2/components/core/Layout/UserInfo/ClaimButton.tsx
+++ b/src/v2/components/core/Layout/UserInfo/ClaimButton.tsx
@@ -10,8 +10,11 @@ const Button = styled(motion.button, {
   padding: "5px 1rem",
   "&:disabled": {
     backgroundColor: "$gray",
+    color: "white",
   },
 });
+
+const transifexTags = "v2/ClaimButton";
 
 interface ClaimButtonProps {
   onClick: () => void;
@@ -19,14 +22,21 @@ interface ClaimButtonProps {
 }
 
 export function ClaimButton({ loading, onClick }: ClaimButtonProps) {
-  const eventListener = useCallback<(e: MouseEvent) => void>((e) => {
-    e.stopPropagation();
-    onClick();
-  }, [onClick])
+  const eventListener = useCallback<(e: MouseEvent) => void>(
+    (e) => {
+      e.stopPropagation();
+      onClick();
+    },
+    [onClick]
+  );
 
   return (
     <Button disabled={loading} onClick={eventListener}>
-      <T _str="Get Rewards" _tags="v2/ClaimButton" />
+      {loading ? (
+        <T _str="Loading" _tags={transifexTags} />
+      ) : (
+        <T _str="Get Rewards" _tags={transifexTags} />
+      )}
     </Button>
   );
 }
--- a/src/v2/components/core/Layout/UserInfo/ClaimButton.tsx
+++ b/src/v2/components/core/Layout/UserInfo/ClaimButton.tsx
@@ -1,11 +1,12 @@
 import { motion } from "framer-motion";
-import React from "react";
+import React, { MouseEvent, useCallback } from "react";
 import { T } from "src/renderer/i18n";
 import { styled } from "src/v2/stitches.config";
 
 const Button = styled(motion.button, {
   appearance: "none",
   backgroundColor: "#dc9c2d",
+  border: "none",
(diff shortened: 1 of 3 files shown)


src/v2/components/core/Layout/UserInfo/ClaimButton.tsx
```diff
@@ -1,26 +1,42 @@
...
+
 interface ClaimButtonProps {
   onClick: () => void;
   loading: boolean;
 }
 
 export function ClaimButton({ loading, onClick }: ClaimButtonProps) {
+  const eventListener = useCallback<(e: MouseEvent) => void>(
+    (e) => {
+      e.stopPropagation();
```
BasixKOR, line 27: This prevents triggering the monster collection by disabling the event bubbling-up, as the parent launches the monster collection when clicked. Check [this website](https://domevents.dev/) to learn more about this.

area363 (approved):

src/v2/views/ClaimCollectionRewardsOverlay/ClaimContent.tsx
```diff
@@ -42,14 +62,14 @@ function ClaimContent({ data, onActionTxId, rewards, tip }: ClaimContentProps) {
   );
 
   useEffect(() => {
-    if (hasMultipleAvatars) return;
+    if (hasMultipleAvatars || !isOpen) return;
```
riemannulus, line 65: What's the intent of this?

BasixKOR: It is a bit complicated. Before the change, the ClaimContent was rendered only when a user clicked the 'Get Rewards' button, and the overlay was open. However, because this behaviour has changed the ClaimContent now renders regardless of 'Get Rewards' has been clicked or not, potentially creating invalid transactions by this `useEffect` statement.

riemannulus: > However, because this behaviour has changed the ClaimContent now renders regardless of 'Get Rewards' has been clicked or not, potentially creating invalid transactions by this useEffect statement.

I have no context about these changes, is it just for V2?

riemannulus: If then, can you explain more about those changes?

BasixKOR: Yes, it is specific to V2 only. The v1 behaviour wouldn't be affected.

riemannulus: I mean, do users not need the claim button (aka `Get Rewards` button) to click anymore in V2? I don't have a context about that.

BasixKOR: When the `OverlayBase` (which is a fancy mechanism for displaying a dialog on V2) is rendered in [`ClaimCollectionRewardsOverlay`](https://github.com/planetarium/9c-launcher/pull/1246/files#diff-9508a74246960570acb5cc90b21479291632541e102d8852defa7e4849a93e62) directly, the component doesn't know whether the user has multiple avatars and showed dialog anyway, which resulted in empty dialog flickering if the user has only one avatar.

However, the `ClaimContent` knows whether the user has multiple avatars, so if it controls whether the dialog should be shown or not, it can hide the dialog when it is unnecessary.

riemannulus: Hmm.. then in this comment

> However, because this behaviour has changed the ClaimContent now renders regardless of 'Get Rewards' has been clicked or not, potentially creating invalid transactions by this useEffect statement.

transactions are just a typo about the transition? I feel like that in my understanding.

BasixKOR: It is kinda both, I forgot to mention it. Because the dialog is now rendered inside ClaimContent the component is rendered regardless of it's opened or not, causing to send tx if there's only one avatar.

riemannulus: Understand

riemannulus (approved):

PR #1249, opened 2022-03-11
Fix V2 Updates

BasixKOR:
- Ensure the listeners are connected correctly.
- Update the visual of UpdateView

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -113,7 +113,7 @@ let remoteHeadless: RemoteHeadless;
 let useRemoteHeadless: boolean;
 let remoteNode: NodeInfo;
 
-const isV2 = app.commandLine.hasSwitch("v2");
+const isV2 = app.commandLine.hasSwitch("v2") || getConfig("UseV2Interface");
 
 ipv4().then((value) => (ip = value));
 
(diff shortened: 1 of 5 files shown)


area363 (approved):

src/main/main.ts
```diff
@@ -219,13 +219,16 @@ async function initializeApp() {
...
+    if (isV2) win = await createV2Window();
     else win = await createWindow();
     createTray(path.join(app.getAppPath(), logoImage));
 
     const u = await checkForUpdates(standalone);
-    if (u) update(u, updateOptions, win);
+    if (u && !isV2) update(u, updateOptions, win);
+    else if (u && isV2)
+      ipcMain.handle("start update", async () => {
+        await update(u, updateOptions, win!);
```
riemannulus, line 230: What is the difference between line 230 and line 227's `update()` method? why do not have `!` on the 227?

BasixKOR: It runs the update process immediately on the V1 launcher as this matches the current behaviour. On the V2 it was difficult to start the update process after the event listeners are initialized correctly, so instead of starting it immediately I changed to let V2 frontend trigger it [here](https://github.com/BasixKOR/9c-launcher/blob/d1978aadab42a365748bf5533f1209e9cc267c7c/src/v2/machines/updateMachine.ts#L48)

BasixKOR: I expect to altogether remove this branching after the V1 launcher sunsetting, but for now, for the backward compatibility, this would do the trick.

riemannulus (approved):

PR #1293, opened 2022-03-29
See tree hashes to check submodules

dahlia:
Now it passes if two submodules share the same tree even though their head commits are not the same.  It is less picky than the previous logic, but still guarantees two submodules share the same source code.

As it relies subprocess calls to `git`, I find bash better for the job.  (Compare `getSubmoduleHash()` with `get-commit-hash`.)  So I rewrote the whole script from TypeScript to bash.

To utilize GitHub Actions' workflow commands, I declared logging functions like `notice`, `warning`, and `error`.  These workflow commands are printed only on GitHub Actions.  On ordinary terminal, these logs are colored instead.  (However, these ANSI escape sequences are stripped if not `isatty()`.)

It looks like below in case two submodules refer to different commits with different trees (note that it terminates with exit code 1):

It looks like below in case two submodules refer to different commits, but they still have the equivalent tree (note that still the exit code is 0 even though it prints yellow warning messages):

It looks like below in case two submodules refer to the exactly same commit (node that the exit code is 0):


scripts/check-submodules.sh
```diff
@@ -0,0 +1,112 @@
...
+  title="$2"
+  if [[ -t 0 ]]; then
+    case "$msg_type" in
+    "error")
+      echo -ne '\033[0;31m'
+      ;;
+    "warning")
+      echo -ne '\033[0;33m'
+      ;;
+    "notice")
```
BasixKOR, line 23: It seems like the "notice" level logs also go to stderr. Is this intended behaviour?

BasixKOR (approved):
LGTM!

longfin (approved):

scripts/check-submodules.sh
dahlia: Yes, it's intended.  I think it's a human-readable diagnostic message, and the script's only output for machine is just the exit code.  (This SO link might be worth to read: <https://unix.stackexchange.com/a/331620/9939>.)

BasixKOR: Ooh! Thanks for the pointer. Glad to learn that :)

PR #1299, opened 2022-03-31
Copy account address when UserInfo item is clicked

BasixKOR:
image
You can copy the address when you click it. And it gives a nice toast when copied!


BasixKOR:
Sadly `react-hot-toast` depends on another CSS-in-JS library called [`goober`](https://github.com/cristianbote/goober) but the footprint is around 1kb.

hskim881028 (approved):

boscohyun (approved):

PR #1308, opened 2022-04-05
Check remote node's APV on ping

### LINKED ISSUE #1260: Update via APV instead of event
Related issue: https://github.com/planetarium/NineChronicles.Headless/issues/1127

BasixKOR:
Fixes #1260.

### DIFF
--- a/src/api.graphql
+++ b/src/api.graphql
@@ -349,6 +349,10 @@ mutation StageTxV2($encodedTx: String!) {
 query PreloadEnded {
   nodeStatus {
     preloadEnded
+    appProtocolVersion {
+      version
+      extra
+    }
   }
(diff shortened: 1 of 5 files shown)


area363 (approved):
👍

https://github.com/planetarium/NineChronicles.Headless/commit/068d1ddbb332e36a7b61cedceca3f5572eff3733

The headless GQL for `appProtocolVersion` is already included in `v100160` so bumping isn't necessarily required.

BasixKOR:
I'm closing this one because:

* This approach doesn't quite solve the problem we have.
* `config.ts` is used on both renderer and main process, and this PR breaks the renderer.

I'll revisit this issue in the next PR.

PR #1313, opened 2022-04-06
Add Yarn 3

### LINKED ISSUE #1154: Move to another package manager
As of now Yarn v1 (a.k.a. classic) has been almost deprecated and doesn't get any new features or improvements. The authors of yarn recommend switching to newer versions of Yarn which can be opted in by individual projects.

As such, we should consider and evaluate other package managers as Yarn v1 is very unlikely to get notable improvements from its development team.

| Alternatives | Notes |
| ------------ | --- |
| Yarn Berry   | Recommended by Yarn devs, very extensive |
| pnpm         | Improved node_modules structure but might not work on our project depending on the dependencies |
| npm           | Comes with Node.js by default. |

BasixKOR:
Pretty much experimental yet. Fixes #1154.

The diff is very big and scary but the majority of them are just lockfile changes.

## Considerations
- [ ] Could we use the `pnpm` linker?
  - P'n'P is impossible due to the incompatibility with Electron.
- [x] How do we make `run-script-os` on Yarn 3 or mimic its behaviour?
  - It worked out of the box.


.yarn/plugins/@yarnpkg/plugin-typescript.cjs
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable */
```
moreal, line 1: When I saw this file with `.yarnrc.yml`, these files are intended, right? Because it seems like generated one.

BasixKOR: Yes, it's safe to assume that everything inside .yarn is generated.

moreal (approved):
It looks good to me! 🚀 The CI speed becomes super fast 🚀

PR #1343, opened 2022-04-14
Display password button on hover if message exists

BasixKOR:
This ensures the message and buttons are not shown together while ensuring users can see their password when they want to.

| Before | After |
| --- | --- |
| image | Animation |

## Considerations that should discussed later
- This solution doesn't give users enough clues to let them know they can toggle password view.

### DIFF
--- a/src/v2/components/ui/TextField.tsx
+++ b/src/v2/components/ui/TextField.tsx
@@ -60,6 +60,36 @@ const Label = styled("label", {
   },
 });
 
+const Side = styled("span", {
+  position: "absolute",
+  top: "50%",
+  right: 20,
+  transform: "translateY(-50%)",
+  display: "flex",
+  alignItems: "flex-end",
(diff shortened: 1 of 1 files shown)


src/v2/components/ui/TextField.tsx
```diff
@@ -60,13 +60,44 @@ const Label = styled("label", {
   },
 });
 
+const Side = styled("span", {
+  position: "absolute",
+  top: "50%",
+  right: 20,
```
longfin, line 66: It seems just replacing in this PR but... how about specifying the unit for this value?

longfin (approved):

src/v2/components/ui/TextField.tsx
BasixKOR: The unit-less values are used quite frequently throughout the source code. Let's discuss this on a separate issue.

PR #1410, opened 2022-05-11
Monster Collection Revamping

BasixKOR:


BasixKOR:
Switching this PR to ready for reviews.

Since this PR is rather long, Here is a summary explaining the PR:

 - This PR adds a new Monster Collection overlay on the V2 interface.
   - Note that this doesn't change the old V1 launcher. I expect to do a complete switch to V2 by the time this get introduced, and because of fundamental differences between them the interoperability looks nearly impossible at my glance.
 - This new interface uses newer Monster Collection feature.
   - Users will be able to migrate to this newer Monster Collection. This will be handled in another PR.

src/v2/components/ui/BareInput.tsx
```diff
@@ -0,0 +1,47 @@
...
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      props.onChange?.(e);
+      setValue(e.target.value.slice(0, props.maxLength));
+    },
+    [props.onChange]
+  );
+
+  return (
+    <BareInputWrapper>
+      <LayoutCalculator aria-hidden>{value}</LayoutCalculator>
```
BasixKOR, line 41: Well, this bit is a bit of magic that I 'borrowed' from CSS-Tricks. Basically, this works by letting the browser calculate the width of input by copying the input's value into an inline element, causing the parent to get resized.

This was necessary because the current amount indicator has to come right after the input and the width of both elements are set dynamically.

* https://css-tricks.com/auto-growing-inputs-textareas/

src/v2/views/MonsterCollectionOverlay/base.tsx
```diff
@@ -0,0 +1,214 @@
...
+import activationButtonHoverImg from "src/v2/resources/collection/button-activation-over.png";
+import disabledButtonImg from "src/v2/resources/collection/button-inactive.png";
+import cancelButtonImg from "src/v2/resources/collection/button-cancel.png";
+import cancelButtonHoverImg from "src/v2/resources/collection/button-cancel-over.png";
+import rewardImg from "src/v2/resources/collection/reward-bg.png";
+import dotImg from "src/v2/resources/collection/dot.png";
+import darkTextBg from "src/v2/resources/collection/dark-text-bg.png";
+import itemBg from "src/v2/resources/collection/item-bg.png";
+
+export const theme = {
```
BasixKOR, line 16: I tried to use [the theme feature of Stitches](https://stitches.dev/docs/theming#add-a-new-theme) but it caused some flickering when the overlay got mounted, so I reverted it to a simple object.

I have some mixed feelings about how this theme object should be organized, but I couldn't come up with something consistent. Some elements don't use this theme object at all and the colour and gradients are different by a bit, making the effort to standardize the tokens bit useless.

src/v2/views/MonsterCollectionOverlay/index.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import "normalize.css";
+
+export default {
+  title: "MonsterCollection/MonsterCollectionContent",
+  component: MonsterCollectionContent,
+};
+
+export const FirstPage = () => (
+  <MonsterCollectionOverlayBase as="main">
```
BasixKOR, line 13: I explicitly opted out from OverlayBase because rendering and debugging dialog style on a Storybook felt quite awful. However, I think we should be able to use `MonsterCollectionOverlay` here since that component now manages the state too.

src/v2/views/MonsterCollectionOverlay/reward.tsx
```diff
@@ -0,0 +1,180 @@
...
+);
+
+const ItemGroupStyled = styled("section", {
+  position: "relative",
+  display: "flex",
+  flexDirection: "column",
+  "& + &": {
+    marginLeft: 40,
+  },
+  "& + &::before": {
```
BasixKOR, line 75: A bit of a hacky way to render dividers between `ItemGroup`s to avoid the complexity of rendering dividers using React. It works quite great and we can use `ItemGroup` without worrying about something like: is this the last component? where should I put the dividers?

But although it is a clever hack indeed you could say it is quite over-engineered and we could have gone with just a bunch of `<Divider />` components.

area363:
@BasixKOR Could you check the failed `circleci` tests? 🙏

area363 (approved):

PR #1437, opened 2022-05-26
Correct `schema.graphql`

moreal:
Since f48a993e9936dce510a794e14b6714a2b820196e, the headless GraphQL type patch was applied. This commit synchronize `schema.graphql` with the schema of the headless.

----

I hope this pull request will resolve the `styles` check failure.

### DIFF
--- a/src/schema.graphql
+++ b/src/schema.graphql
@@ -462,7 +462,7 @@ type StakeRegularRewardRowType {
 }
 
 type StakeRegularRewardSheetType {
-  orderedList: [StakeRegularRewardRowType]
+  orderedList: [StakeRegularRewardRowType!]!
 }
 
 type StakeStateType {


BasixKOR:
Has this patch been reflected on the GraphTypes of the headless as well? We typically update the schema by fetching it from the `/schema.graphql` endpoint of a running headless instance, so I think this change needs to be done with headless changes.

Note: the schema updating procedure is largely undocumented for now. I'll add them to the relevant docs when I have time.

moreal:
> Has this patch been reflected on the GraphTypes of the headless as well? We typically update the schema by fetching it from the `/schema.graphql` endpoint of a running headless instance, so I think this change needs to be done with headless changes.

Yes. When you look current [2022q2 branch's NineChronicles.Headless submodule HEAD](https://github.com/planetarium/NineChronicles.Headless/commits/7015c982b0c569fc10c172844d6e83f598b0c9e1), you can see https://github.com/planetarium/NineChronicles.Headless/commit/459a0f3d1685743e0f80abdf1d253a58538271f8 commit to correct the GraphQL type. The commit was applied by bumping the headless submodule in https://github.com/planetarium/9c-launcher/commit/f48a993e9936dce510a794e14b6714a2b820196e commit. I patched the line by copying from the `/schema.graphql` response of local running instance.


> Note: the schema updating procedure is largely undocumented for now. I'll add them to the relevant docs when I have time.

Thank you very much 🙇🏻‍♂️

BasixKOR (approved):

PR #1441, opened 2022-05-26
Use normal weight on reward titles

BasixKOR:
Just a stylistic change.


moreal (approved):

BasixKOR:
Merging since I don't expect any major regression from a stylistic change.

PR #1442, opened 2022-05-26
Change staking menu item to newer staking

BasixKOR:


area363 (approved):

moreal (approved):

src/v2/components/core/Menu/index.tsx
```diff
@@ -32,19 +33,19 @@ const MenuDivider = styled("hr", {
 
 const app = require("electron").remote.app;
 
+type Overlay = "settings" | "staking";
+
 function Menu() {
   const account = useStore("account");
-  const [isSettingsOpen, setIsSettingsOpen] = useState(false);
+  const [currentOverlay, openOverlay] = useState<Overlay | null>(false);
```
moreal, line 40: It seems to need to update the default value as `null`.

```suggestion
  const [currentOverlay, openOverlay] = useState<Overlay | null>(null);
```

moreal (changes requested):

moreal (approved):
🙇🏻‍♂️

PR #1449, opened 2022-05-27
Fix command to staking

BasixKOR:
Includes genesis block hash change. This commit c79b284 should be reverted before releasing this feature to mainnet.

### DIFF
--- a/src/main/headless/action.ts
+++ b/src/main/headless/action.ts
@@ -69,7 +69,7 @@ export class Action extends StandaloneSubcommand {
 
   public Stake(amount: string, path: string): boolean {
     try {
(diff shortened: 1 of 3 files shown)


area363 (commented):
Could you include a comment in the headless.ts that the genesis hash should be reverted when merging previewnet to the main/dev branch? I think it's highly likely that we forget 😅

BasixKOR:
Yeah. It should be fetched dynamically tbh. It seems like it can be fetched via `query.nodeStatus.genesis.hash` so we should be moving to query them instead of hard-coding them. I'll track this on a separate issue.

area363 (approved):

PR #1483, opened 2022-06-03
Merge 2022q2 into previewnet

BasixKOR:
Rebased 2022q2 due to extra commits on previewnet.


moreal (approved):

area363:
https://app.circleci.com/pipelines/github/planetarium/9c-launcher/1140/workflows/9ecc7fc3-0fef-4c09-9a7e-bd05af97440a/jobs/4114?invite=true#step-112-544
Some e2e tests are failing 😢

BasixKOR:
Failing E2E test seems like a false positive rather than an actual issue. I was able to get the launcher up and running just fine. Although this failure needs an investigation of its own but it shouldn't be a blocking issue for now.

PR #1487, opened 2022-06-03
Disable level if amount is insufficient

BasixKOR:
image

It's gonna look like this.


src/v2/views/MonsterCollectionOverlay/level.tsx
```diff
@@ -119,15 +124,21 @@ interface LevelProps {
    * It is meant to be used on editing.
    */
   selected?: boolean;
+  /**
+   * Whether if it's avauilable to be chosen or not.
```
moreal, line 128: Just a simple typo.

```suggestion
   * Whether if it's available to be chosen or not.
```

PR #1504, opened 2022-06-08
Implement migration from monster collection V1

BasixKOR:


src/v2/views/MonsterCollectionOverlay/Migration.tsx
```diff
@@ -0,0 +1,87 @@
...
+  );
+  const elapsedBlocks = tip - collectionState.startedBlockIndex;
+  const elapsed = getRemain(elapsedBlocks);
+
+  return (
+    <MigrationAlert isOpen={true} onConfirm={open} isClaimable={isClaimable}>
+      <MigrationAlertItem title="Deposit amount">
+        <strong>
+          <img src={ncgIcon} />
+          {deposit}
```
BasixKOR, line 55: Will break if collectionSheet is unavailable, or there's no matching level in the collectionSheet.

src/v2/views/MonsterCollectionOverlay/index.tsx
```diff
@@ -66,9 +71,21 @@ function MonsterCollectionOverlay({ isOpen, onClose }: OverlayProps) {
...
-      />
+      >
+        {collection.stateQuery.monsterCollectionState && (
+          <Migration
+            tip={tip.nodeStatus.tip.index}
+            collectionState={collection.stateQuery.monsterCollectionState}
+            collectionSheet={collection.stateQuery.monsterCollectionSheet}
+            onActionTxId={(txId) => {
+              setLoading(true);
+              fetchStatus({ variables: { txId } });
```
BasixKOR, line 84: We're using the same refetching code when we're changing the deposit amount since what it does is refech when the tx is successfully staged.

This allows us to use the same loading screen made for changing the amount.

src/v2/views/MonsterCollectionOverlay/Migration.tsx
BasixKOR: Maybe we should show the level if it's unavailable?

PR #1508, opened 2022-06-09
Add onClose handler to Migration

BasixKOR:


src/v2/views/MonsterCollectionOverlay/Migration.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useMemo, useReducer } from "react";
+import React, { useEffect, useMemo, useReducer } from "react";
```
moreal, line 1: I'm not sure but I can't find where the `useEffect` was used 🤔

moreal (approved):

src/v2/views/MonsterCollectionOverlay/Migration.tsx
BasixKOR: Ooh yeah I was doing some experiments while I was working on this 😂 It is a unused import.

PR #1509, opened 2022-06-09
Hide remaining days if it's claimable

BasixKOR:


src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
```diff
@@ -228,12 +228,15 @@ export function MonsterCollectionContent({
...
-        {!isEditing && stakeState && tip && (
-          <DepositDescription>
-            About {getRemain(stakeState.claimableBlockIndex - tip)} of deposit
-            days!
-          </DepositDescription>
-        )}
+        {!isEditing &&
+          stakeState &&
+          tip &&
+          tip > stakeState.claimableBlockIndex && (
```
moreal, line 234: I'm not sure about JavaScript language but in this expression, the tip should be less than `claimableBlockIndex`, to hide the remaining days if it's claimable now.

When the tip is more than `claimableBlockIndex`, it means it's claimable so I thought the component will be shown. `stakeState.claimableBlockIndex - tip` will be always negative.

image

BasixKOR: You're right it's the opposite

moreal (approved):
👍🏻

PR #1510, opened 2022-06-09
Look up bonusRewards for display

BasixKOR:
This won't work unless the headless is bumped. Thus untested.


BasixKOR:
It looks like the diff of `schema.graphql` is somewhat broken? I'm guessing it's CRLF issues, but the only changed lines are between 458 to 467.

src/schema.graphql
```diff
@@ -1,649 +1,655 @@
...
-}
-
-type StakeRegularRewardRowType {
-  level: Int!
-  requiredGold: Long!
-  rewards: [StakeRegularRewardInfoType!]!
-}
-
-type StakeRegularRewardSheetType {
+schema {
```
moreal, line 1: It isn't a suggestion to do in this pull request. But it seems we should configure the line ending rule with editorconfig or else. In my memory, the change of all lines appeared sometimes and it was because of the line ending (CRLF -> LF or LF -> CRLF).

moreal (approved):

PR #1528, opened 2022-06-14
Fix bonusRewardMap deps

BasixKOR:
The bonusRewardMap was not updated correctly when the amount has changed, so this PR aims to fix this.


longfin (approved):

moreal (approved):

longfin:
But well... from my point of view (don't having context), it seems be just a simple refactoring. 😕

moreal (commented):
I'm not sure but can it depend on `bonusRewards` directly?

longfin:
I asked @BasixKOR directly, and he answered that, injecting `index` to `useMemo()` is matter. (`levels` is considered like sort of constant 😅 )

PR #1541, opened 2022-06-15
Track staking events on Mixpanel

BasixKOR:
Currently emits:
 - **Staking/Claim**: An avatar address, and TxId.
 - **Staking/AmountChange**: A new amount and a previous amount.
 - **Staking/Migration**: A block index when the Tx initiated (inaccurate), TxId, and a level of migrated state.

### DIFF
--- a/src/v2/views/MonsterCollectionOverlay/index.tsx
+++ b/src/v2/views/MonsterCollectionOverlay/index.tsx
@@ -95,7 +95,7 @@ function MonsterCollectionOverlay({ isOpen, onClose }: OverlayProps) {
               ipcRenderer.send("mixpanel-track-event", "Staking/Migration", {
                 txId,
                 tip: tip.nodeStatus.tip.index,
-                amount: collection.stateQuery.monsterCollectionState?.level,
+                level: collection.stateQuery.monsterCollectionState?.level,
               });
               fetchStatus({ variables: { txId } });
             }}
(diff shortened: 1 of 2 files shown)


src/v2/views/MonsterCollectionOverlay/index.tsx
```diff
@@ -87,6 +92,11 @@ function MonsterCollectionOverlay({ isOpen, onClose }: OverlayProps) {
             collectionSheet={collection.stateQuery.monsterCollectionSheet}
             onActionTxId={(txId) => {
               setLoading(true);
+              ipcRenderer.send("mixpanel-track-event", "Staking/Migration", {
+                txId,
+                tip: tip.nodeStatus.tip.index,
+                amount: collection.stateQuery.monsterCollectionState?.level,
```
moreal, line 98: If the `amount` was intended to represent its deposit, it seems not to represent it correctly. In my thought, the `level` naming is enough to represent the amount. The `level`s can be converted to `amount`.

Maybe it needs `amount` when the Mixpanel user wants to summarize them for the total migrated amount. But then, it needs to fetch `query.stateQuery.monsterCollectionState.address`' deposit from `query.goldBalance(address)`.

moreal: I'm sorry that I didn't see the PR description 😢 The above comment is for only the naming.

BasixKOR: Oh right, I tried to convert the level to the amount by looking it up on the sheet but I haven't really done them. I'll rename it to level.

moreal (approved):
It seems good to me but it will be good to receive approvals from the Mixpanel users.

PR #1555, opened 2022-06-16
Pad private key when passing it to the headless

BasixKOR:
This fixes some users' issues with private keys shorter than 32 bytes.


area363 (approved):
👍

Issue #1565, opened 2022-06-17
Improve ESLint rules and actually use them


BasixKOR:
Brutally honestly we haven't really run lint for a very long time and the rules are outdated compared to what we're actually writing. We've met with a couple of production errors that could've been prevented with simple lint (`react-hooks/exhaustive-dep` would prevent 9/10 of those) and it's having a negative impact on the codebase as a whole.

What we need to do:
 - [x] Update rules
   - I'd like to remove airbnb rules since they don't really match what we're actually writing
 - [x] Fix the lint errors
 - [x] Add CI jobs to run them

Akamig:
I think we would agree on a gradual and minimal-to-complicated approach to those rules since lint has been not used for a long time, and there are about 713 problems according to the current ruleset solely on V2 source code. So, what would be a nice starting point?

BasixKOR:
We need to do some evaluations on the known configurations but I think [CRA's default]https://github.com/facebook/create-react-app/tree/master/packages/eslint-config-react-app) looks reasonable.

PR #1566, opened 2022-06-20
Applying Lint Globally

Akamig:
This addresses #1565.
Alright. so currently I used these ruleset,
```
{
  "parser": "@typescript-eslint/parser",
  "plugins": ["@typescript-eslint"],
  "extends": [
    "react-app",
    "eslint:recommended",
    "plugin:@typescript-eslint/recommended",
    "plugin:prettier/recommended"
  ],
  "rules": {
    "jsx-a11y/alt-text": 0,
    "react/prop-types": 0,
    "no-underscore-dangle": 0,
    "import/imports-first": ["error", "absolute-first"], <- focus on this later
    "import/newline-after-import": "error",
    "import/prefer-default-export": 0,
    "semi": "error",
    "require-await": "error"
  },
  "globals": {
    "window": true,
    "document": true,
    "localStorage": true,
    "FormData": true,
    "FileReader": true,
    "Blob": true,
    "navigator": true,
    "Headers": true,
    "Request": true,
    "fetch": true
  }
}
```
Then got stastistic out most frequent errors and warnings by this ruleset with eslint-stats, and here's the result.
```
By Warnings

jsx-a11y/anchor-is-valid:                    1|
no-loop-func:                                1|
no-sequences:                                1|
import/no-anonymous-default-export:          2|
no-useless-computed-key:                     2|
react/jsx-no-target-blank:                   2|
@typescript-eslint/no-useless-constructor:   2|
no-useless-concat:                           3|
array-callback-return:                       5|
eqeqeq:                                     16|
@typescript-eslint/no-non-null-assertion:   36|
react-hooks/exhaustive-deps:                37|
@typescript-eslint/no-unused-vars:         101|
@typescript-eslint/no-explicit-any:        368|

By Errors

no-fallthrough:                             1|
no-restricted-globals:                      1|
react-hooks/rules-of-hooks:                 1|
semi:                                       1|
@typescript-eslint/no-empty-interface:      1|
no-prototype-builtins:                      2|
no-constant-condition:                      3|
@typescript-eslint/no-unused-expressions:   3|
@typescript-eslint/ban-ts-comment:          4|
no-var:                                     5|
@typescript-eslint/no-var-requires:         6|
@typescript-eslint/no-empty-function:      15|
@typescript-eslint/no-inferrable-types:    24|
require-await:                             37|
prefer-const:                              40|
@typescript-eslint/ban-types:             161|
import/imports-first:                     290|
```
From this result, first, ~we just can ignore no-explicit-any~ I personally think "import/imports-first" rule can be replaced with [import/first](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/imports-first.md), instead of [import/order](https://github.com/import-js/eslint-plugin-import/blob/376747914b47fbdcf99212b9e9bd4d5e09825385/docs/rules/order.md) which enforces import order for absolute path modules first. frankly, I don't think import order according to its path doesn't matter much enough to be considered as an error. as long as imports are at the top.

image
secondly, I found edge case "semi" rule fighting with prettier (/src/interfaces/i18n.ts exactly), then I found out prettier runs inside of eslint, and while It doesn't say 'not recommended', [Prettier documentation sounds like 'you'd better run just prettier --write or --check outside the lint'](https://prettier.io/docs/en/integrating-with-linters.html).
Can I,
- Delete "semi", "require-await", "import/newline-after-import"
- Seperate Prettier from ESLint and for the long run, separately run both from CI?


BasixKOR:
`plugin:prettier/recommended` is intended to disable any rules that conflicts with Prettier, so I think you can safely remove any conflicting ones.

Akamig:
> `plugin:prettier/recommended` is intended to disable any rules that conflict with Prettier, so I think you can safely remove any conflicting ones.

Yep, it does, but I prefer to use only `eslint-config-prettier` to disable eslint rules for prettier and run prettier seperately,
And not use `eslint-plugin-prettier` which runs prettier rules inside eslint and outputs errors for minor style issues (`prettier/prettier: error` in `plugin:prettier/recommended`). [Prettier documentation](https://prettier.io/docs/en/integrating-with-linters.html#notes) also addresses this as:

> These plugins were especially useful when Prettier 'was' new. 

> The downsides of those plugins are:
> - You end up with a lot of red squiggly lines in your editor, which gets annoying. Prettier is supposed to make you forget about formatting – and not be in your face about it!
> - They are slower than running Prettier directly.
> - They’re yet one layer of indirection where things may break.

and I agree with it.

BasixKOR:
Oops I didn't see it was a plugin, I blindly assumed that it would be a configuration 😂 I agree with your concerns.

BasixKOR:
Could you commit `yarn.lock` file as well? I think it's causing CI to fail.

Akamig:
Now, I applied some simple fixes on v2 (no-require), and disabled some rules, mostly from typescript-eslint,
Justifications here.
`@typescript-eslint/no-inferrable-types`
I don't think notation for simply inferrable types are redundunt. ~~I don't believe TS type inferrer~~

`@typescript-eslint/no-explicit-any`
~~Who DARES try to use Typescript without any~~, I mean, seriously. 

`@typescript-eslint/no-unused-vars`
I read through most of the codes, and it seems mostly for ease of autocomplete. pretty harmless for code.

`@typescript-eslint/ban-types, disabled for {}`
I think using an empty object literal as `{}` is cleaner, simpler and faster, and honestly faster to acknowledge. I disagree with this rule.

So with this ruleset, we currently have about 30 errors for whole codebase, here's eslint-stats result by error for whole ./src
```
✖ 136 problems (30 errors, 106 warnings)
no-fallthrough:                            1|
no-restricted-globals:                     1|
react-hooks/rules-of-hooks:                1|
@typescript-eslint/no-var-requires:        1|
no-prototype-builtins:                     2|
@typescript-eslint/ban-types:              2|
no-constant-condition:                     3|
@typescript-eslint/no-unused-expressions:  3|
@typescript-eslint/ban-ts-comment:         4|
@typescript-eslint/no-empty-function:     12|
```
I think these 'errors' (mostly not, for example, `no-empty-function` is caused by noop functions.) can be addressed manually. and 106 warnings are mostly ignorable. I'll left the rest for the maintainer's taste of linter setting.

BasixKOR:
I'm just guessing that most of them can be addressed later. Nice work!

BasixKOR (approved):
Yeah looks OK to me. I'll take care of the lockfile.

PR #1570, opened 2022-06-20
Implement difference viewer on collection

BasixKOR:
image

### DIFF
--- a/src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
+++ b/src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
@@ -75,6 +75,36 @@ const images = [
   monster5Img,
 ];
 
+type Levels =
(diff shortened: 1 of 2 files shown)


src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
```diff
@@ -74,6 +74,36 @@ const images = [
   monster5Img,
 ];
 
+type LevelList =
```
Akamig, line 77: https://github.com/typescript-eslint/typescript-eslint/issues/2585#issuecomment-696237148

> Note: Even with this option set to true, this rule will report if you name a type and a variable the same name. This is intentional. Declaring a variable and a type and a variable the same is usually an accident, and it can lead to hard-to-understand code.

Bit hard to agree with, but well. there's a point.

Akamig (approved):

PR #1573, opened 2022-06-22
Prevent depositing same amount as before

BasixKOR:


moreal (approved):
Though changed lines seem many, the original implementation becomes indented as one step and I understood it to be throttled so that even if multiple requests come in, only one is executed at a time, with the `inProgress` boolean flag.

If I understood well, it seems good 👍🏻

Issue #1595, opened 2022-06-23
Support Linux


Akamig:
period.

I'm working on it.

PR #1598, opened 2022-06-23
Support Linux

Akamig:
image

#1595 
Running on linux confirmed.


BasixKOR:
It should be noted that currently, the game itself doesn't support Linux and any attempts to actually run them will fail. Nice work though.

Akamig:
Then I'm going to dig in some CI codes in GH Workflows on the 9c side, officially build for Linux, then I'll comeback here.

Issue #1158, opened 2022-01-14
Move macOS build to Circle CI


BasixKOR:
We've moved the Windows build process to Circle CI for improved performance and better management. We're quite satisfied with the result but we haven't moved macOS ones due to the lower priority. Ideally, we should move them to Circle CI as well, and preferably do the common tasks within the separated workflow.

Akamig:
YAY

PR #1664, opened 2022-07-11
Move macOS/Linux build to Circle CI

Akamig:
This includes resolve of #1158 

- have issue with signing/making apv in some OS now
- window build fails for some reason (pointing node.js version for now)
- successfully building with macOS and Linux


longfin (approved):

.circleci/config.yml
```diff
@@ -38,67 +50,127 @@ commands:
...
+                  curl "https://awscli.amazonaws.com/AWSCLIV2.pkg" -o "AWSCLIV2.pkg"
+                  sudo installer -pkg AWSCLIV2.pkg -target /
+                name: Install .NET Core, AWS
+      - when:
+          condition:
+            equal: [ *linux, << parameters.os >>]
+          steps:
+            - run: 
+                command: |
+                  wget https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
```
BasixKOR, line 90: Using [`dotnet-install` script](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) could be more beneficial in this case since we don't need to have a separate macOS/Linux installation process.

.circleci/config.yml
```diff
@@ -1,8 +1,19 @@
...
 
 orbs:
-  node: circleci/node@4.7
-  win: circleci/windows@2.4.1
+  win: circleci/windows@4.1.1
+
+executors:
+  windows: &windows
+    machine:
+      image: windows-server-2022-gui:edge
```
BasixKOR, line 9: Why are we using the `edge` variant?

BasixKOR: Also CircleCI strongly recommend using orbs instead of configuring machine directly, so I think we should stick to orbs. https://circleci.com/docs/using-windows

.circleci/config.yml
```diff
@@ -38,67 +51,127 @@ commands:
...
           working_directory: NineChronicles.Headless
           name: Clean Headless
       - run:
-          command: yarn concurrently npm:build-headless npm:bundle-player
-          shell: bash
+          command: yarn concurrently yarn:build-headless yarn:bundle-player
           name: Setup Software Components
-      - run:
+      - run: 
+          command: 'echo "::set-output name=apv::$(scripts/make-apv.sh)"'
```
BasixKOR, line 106:
```suggestion
      - run: 'echo "::set-output name=apv::$(scripts/make-apv.sh)"'
```

You can write like this if the step only has a command. You may want to name this though.

.circleci/config.yml
```diff
@@ -38,67 +51,127 @@ commands:
...
+            equal: [ *macos, << parameters.os >>]
+          steps:
+            - run: 
+                command: aws s3 cp pack-dist/macOS.tar.gz s3://9c-artifacts/9c-launcher/$CIRCLE_SHA1/macOS.tar.gz
+      - when:
+          condition:
+            equal: [ *linux, << parameters.os >>]
+          steps:
+            - run: 
+                command: aws s3 cp pack-dist/Linux.tar.gz s3://9c-artifacts/9c-launcher/$CIRCLE_SHA1/Linux.tar.gz
```
BasixKOR, line 173: Could we simplify this by any chance?

.circleci/config.yml
```diff
@@ -155,11 +228,16 @@ jobs:
...
+              os: ["linux", "macos", "windows"]
       - styles
       - update-translations:
           filters:
             branches:
               only: development
       - e2e-test
       - chromatic
+
+
```
BasixKOR, line 242: Extra lines.

.circleci/config.yml
```diff
@@ -38,67 +51,127 @@ commands:
...
+            - run: 
+                command: |
+                  chmod +x resources/app/9c
+                  tar cvfz ../../pack-dist/Linux.tar.gz *
+                working_directory: pack/Nine Chronicles-linux-x64/
       - store_artifacts:
           path: pack-dist
           destination: dist-artifacts
-      - run:
+      - run:
```
BasixKOR, line 151: Extra space. Probably running Prettier once would solve most of the styling issues.

.circleci/config.yml
Akamig: For Using Orbs, although it is technically possible to use orbs with matrix jobs, but because we can't set parameter names for orbs (i.e. &windows, &macos), matrix itself and OS conditional statements really gets ugly. so I had to use machine directly. https://circleci.com/docs/using-matrix-jobs

Akamig: And for the `edge` variant, If I remember correctly `current` variant's preinstalled .NET SDK version stays on 5, so In order to utilize preinstalled SDK to minimize installation time, I used `edge` variant. 

But on the other hand, I'm not sure about it, so I'll try the `current` variant shortly after.

BasixKOR: Good news for ya, .NET 6 is available on `current` as well: https://discuss.circleci.com/t/june-2022-windows-updates-available-for-current-tags/44540

.circleci/config.yml
Akamig: [Sadly there's no `else if` in circleci's logic statements](https://circleci.com/docs/configuration-reference#logic-statements), so I'm pretty sure this is best we can do for now.

.circleci/config.yml
Akamig: Added Name on it.

.circleci/config.yml
Akamig: Um, we actually need to have separate macOS/Linux installation process, because we have AWS CLI integrated on ubuntu image, but there's no AWS or .NET SDK integrated in macOS image. so, for the ease of process (i.e. it is separated after all, and dotnet-install sometimes didn't worked for some reason) I used each package managers to install .NET SDK.

.circleci/config.yml
Akamig: Confirmed, using `current`.

PR #1670, opened 2022-07-14
The Great Electron Bump

BasixKOR:
We're bumping Electron to the latest version available. This PR aims to fix any incompatible code and migrates it.

Due to the migration cost, it relies heavily on `@electron/remote`. This is a debt, and we'll migrate later on.


Akamig:
As I tested there was no severe problem, and I think the great electron bump is completed now, and It'd be better to merge this and make separate PR for e2e test revise. what do you think?

BasixKOR:
This PR will be deferred from merge until we figure out the version to release this.

Akamig (approved):
Looks awesome to me!

PR #1676, opened 2022-07-20
Implement CompatiblityVersion check

BasixKOR:
In this PR, we're proposing a new APV extra that prevents the launcher from auto-updating itself and asking users to do something else; this allows us to deliver updates that weren't possible due to the updater's limitations.

### DIFF
--- a/src/main/update.ts
+++ b/src/main/update.ts
@@ -1,5 +1,5 @@
 import { encode, decode, BencodexDict } from "bencodex";
-import { DownloadItem, app } from "electron";
+import { DownloadItem, app, dialog, shell } from "electron";
 import { download, Options as ElectronDLOptions } from "electron-dl";
 import extractZip from "extract-zip";
 import * as utils from "../utils";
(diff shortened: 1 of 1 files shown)


src/main/update.ts
```diff
@@ -122,6 +132,37 @@ export async function update(update: Update, listeners: IUpdateOptions) {
...
+  if (compatVersion > currentCompatVersion) {
+    console.log(
+      `Stop update process. CompatiblityVersion is higher than current.`
+    );
+    win?.webContents.send("compatiblity-version-higher-than-current");
+    if (win) {
+      const { checkboxChecked } = await dialog.showMessageBox(win, {
+        type: "error",
+        message: t(
+          "This update needs reinstallation due to techincal issues.",
```
Akamig, line 151: rather than "This update needs reinstallation due to technical issues", "This version of launcher is no longer supported and requires force re-installation due to technical issues." sounds more coercing.

longfin (approved):

PR #1677, opened 2022-07-20
Replace Spectron with Playwright

BasixKOR:
Prerequisite: #1670.
Fixes: #1678.

This PR replaces Spectron with [Playwright's experimental Electron support](https://playwright.dev/docs/api/class-electron).


Akamig (approved):
LGTM

PR #1687, opened 2022-07-25
Get rid of all eslint errors and add lint in CI job

Akamig:
This partiallty resolves second TODO of #1565.

### rule justification
- "@typescript-eslint/no-empty-function"
All of use cases of empty functions were for preventing default action of certain callbacks or events.
eslint suggests leave explicit` //this is empty function` commentation, but I think we can just ignore it.
- no-constant-condition
It's needless to say that `while(true)` is more coherent than `while (typeof x === "undefined")` which [eslint suggests](https://eslint.org/docs/latest/rules/no-constant-condition) as alternative.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -187,6 +187,9 @@ jobs:
       - run:
           command: yarn tsc --noEmit
           name: Type Check
+      - run:
+          command: yarn eslint ./src --quiet
+          name: ESLint Check
   chromatic:
     docker:
       - image: cimg/node:lts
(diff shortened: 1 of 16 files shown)


BasixKOR (approved):
We could probably use lodash/noop or something, but I don't think it's worth the cost either. Good work!

Akamig:
With [5313ebe](https://github.com/planetarium/9c-launcher/pull/1687/commits/5313ebe1182ddb106e9bbccbcbb19dc387ca4ebf), I added ESLint error checking in CI jobs. which resolves third TODO of #1565. which seal the deal.

PR #1712, opened 2022-08-02
Run setup commands in postinstall 

BasixKOR:
This PR includes common setup commands into the `postinstall` script for greatly improved usability.


BasixKOR:
Closing as most steps of this PR is heading to be replaced.

Issue #1707, opened 2022-08-01
Tune updater


longfin:

Atralupus:
This issue is almost same [1705](https://github.com/planetarium/9c-launcher/issues/1705).

Issue #1706, opened 2022-08-01
Design version scheme for client


longfin:

Atralupus:
Add `WindowsPlayerBinaryUrl`, `macOSPlayerBinaryUrl` in apv extra data

Atralupus:
https://{domain}/{rc version}/{project}/{binary}

i.e
https://download.ninechronicles.com/v100270/player/win.zip
https://download.ninechronicles.com/v100270/player/mac.tar.gz
https://download.ninechronicles.com/v100270/launcher/win.zip
https://download.ninechronicles.com/v100270/launcher/mac.tar.gz

BasixKOR:
You may want to consider adding architecture types to the URL as well. You know, future-proofing.

Atralupus:
umm... Player, Headless, etc... can be compatible with multiple rc versions.

https://{domain}/{project}/{project version}/{rc version}/{architecture}

i.e
~/player/v100001/rc100270/win.zip
~/player/v100001/rc100271/arm64
~/launcher/v100003/rc100270/win.zip

Atralupus:
> umm... Player, Headless, etc... can be compatible with multiple rc versions.
> 
> https://{domain}/{project}/{project version}/{rc version}/{architecture}
> 
> i.e ~/player/v100001/rc100270/win.zip ~/player/v100001/rc100271/arm64 ~/launcher/v100003/rc100270/win.zip

It's a bad idea... forget it

Atralupus:
https://{domain}/{rc version}/{project}/{project version}/{architecture}

i.e. ~/v100270/launcher/v1/mac.tar.gz

Then we could delete 'WindowsBinaryUrl', 'macOSBinaryUrl' from apv extra data

Atralupus:
https://{domain}/{env}/{rc version}/{project}/{project version}/{architecture}

i.e. ~/internal/v1085/launcher/v1/mac.tar.gz

for internal, previewnet, etc...

Issue #1705, opened 2022-08-01
Implment basic feature for downloading client


longfin:
Implement the 9c client(player) updater to separate the launcher and the client
`downloadPlayer.ts` script already exists for similar operations
This script can be easily resolved by leveraging and implementing it

Atralupus:
[Implementation completed](https://github.com/planetarium/9c-launcher/pull/1727) but... not yet merged into development

PR #1727, opened 2022-08-08
[Feature] 9c Player Download & Update

Atralupus:
Related to https://github.com/planetarium/9c-launcher/issues/1705
Related to https://github.com/planetarium/9c-launcher/issues/1707
Related to https://github.com/planetarium/9c-launcher/issues/1706

This PR will change the 9c player to install in the path '[userData](https://www.electronjs.org/docs/latest/api/app#appgetpathname)/player'

 - [x] Add `WindowsPlayerBinaryUrl`, `macOSPlayerBinaryUrl` in apv extra data
 - [x] Split launcher update, player update
 - [x] Implement 9c player download
 - [x] Replace update script
 - [x] Replace player execute code
 - [x] Fix lint...


Atralupus:
Important thing is the scripts under the `update/*`

src/config.ts
```diff
@@ -178,7 +178,10 @@ export function getBlockChainStorePath(): string {
 export const REQUIRED_DISK_SPACE = 20n * 1000n * 1000n * 1000n;
 export const SNAPSHOT_SAVE_PATH = app.getPath("userData");
 export const MAC_GAME_PATH = "9c.app/Contents/MacOS/9c";
```
BasixKOR, line 180: This path should be edited too.

src/main/update/extract.ts
```diff
@@ -0,0 +1,57 @@
+import extractZip from "extract-zip";
```
BasixKOR, line 1: Maybe we should consider some generic library for this: https://www.npmjs.com/package/decompress. I guess we can discuss this in another isssue.

src/main/update/extract.ts
```diff
@@ -0,0 +1,57 @@
...
+    dir: tempDir,
+    onEntry: (_, zipfile) => {
+      const progress = zipfile.entriesRead / zipfile.entryCount;
+      win?.webContents.send("update extract progress", progress);
+    },
+  });
+  win.webContents.send("update extract complete");
+  console.log("[player] The zip archive", from, "has extracted to", tempDir);
+  win.webContents.send("update copying progress");
+  await utils.copyDir(tempDir, to);
```
BasixKOR, line 27: Is there a reason not to extract the target directory directly?

src/main/update/player-update.ts
```diff
@@ -0,0 +1,62 @@
...
+  listeners: IUpdateOptions
+) {
+  const win = listeners.getWindow();
+
+  if (win === null) {
+    console.log("Stop update process because win is null.");
+    return;
+  }
+
+  win?.webContents.send("update download started");
```
BasixKOR, line 23: I don't feel okay with reusing the update events for player updates again. As far as my knowledge of the update events, the current `updateMachine.ts` assumes all the steps are irreversible (i.e. you can't go back from extract to download)

src/config.ts
Atralupus: umm... my mistake😞

src/main/update/extract.ts
Atralupus: In the first step, I did try to use the existing scripts
As you said, we should need to create a separate issue

src/main/update/extract.ts
Atralupus: Maybe... Nope.
But as I mentioned in this [comment](https://github.com/planetarium/9c-launcher/pull/1727#discussion_r940002597), I tried to use the existing scripts, So I'll try to solve it another PR

src/main/update/player-update.ts
Atralupus: Thanks you, then... I'll solve to combine window send event

BasixKOR (approved):
Nitpicks included but seems fine.

PR #1731, opened 2022-08-10
[Feature] Separate download & update player from launcher update

Atralupus:
Solved #1704

# Description
All tasks for `Separate player` are merged in this huge PR.

**First major update** is `Doesn't import downloadUrl from APV version`
The `DownloadURL` is created according to the version scheme, and the `DownloadURL` attempts to update it assuming it exists.

**Secondly major update** is you will no longer install the player in the same folder as Launcher.
The update script has been detached to be installed separately in the [userData folder](https://www.electronjs.org/docs/latest/api/app#appgetpathname) and accordingly, modified player execute path.

And... **minor updates**
- If there is no player under the userData folder for the first time, start `player update`
- Not proceeding bundle-player in CI.

This PR will not be merged until the [Internal test](https://github.com/planetarium/9c-launcher/issues/1760) is complete.

### TODO
- [x] Separate update player
- [x] Download URL problem
- [x] ~~Remove 9c submodule~~
- [x] Fix circle ci
- [x] Update downloadPlayer
---
Merged PR
- https://github.com/planetarium/9c-launcher/pull/1727
- https://github.com/planetarium/9c-launcher/pull/1738
- https://github.com/planetarium/9c-launcher/pull/1745
- https://github.com/planetarium/9c-launcher/pull/1761
- https://github.com/planetarium/9c-launcher/pull/1767
- https://github.com/planetarium/9c-launcher/pull/1784

PR #1738, opened 2022-08-11
[Refactor] Doesn't import downloadUrl from apv extra data

Atralupus:
Related to https://github.com/planetarium/9c-launcher/issues/1706, https://github.com/planetarium/9c-launcher/pull/1731

Now, We don't import downloadUrl from apv extra data.
The downloadURL is created according to the version scheme, and the downloadURL attempts to update it assuming it exists.

### DIFF
--- a/src/main/constants.ts
+++ b/src/main/constants.ts
@@ -1,0 +1,1 @@
+export const DOWNLOAD_URL = "https://download.nine-chronicles.com"
--- a/src/main/constants.ts
+++ b/src/main/constants.ts
@@ -1,1 +1,0 @@
-export const DOWNLOAD_URL = "https://download.nine-chronicles.com";
--- a/src/main/constants.ts
+++ b/src/main/constants.ts
@@ -1,1 +1,1 @@
-export const DOWNLOAD_URL = "https://download.nine-chronicles.com"
+export const DOWNLOAD_URL = "https://download.nine-chronicles.com";
--- a/src/main/update/launcher-update.ts
+++ b/src/main/update/launcher-update.ts
@@ -108,16 +108,10 @@ export async function update(update: Update, listeners: IUpdateOptions) {
   const extra = decode(buffer) as BencodexDict;
   console.log("peerVersionExtra (decoded):", JSON.stringify(extra)); // Stringifies the JSON for extra clarity in the log
 
-  // FIXME: project version number hard coding: 1.
   const network = getConfig("Network", "9c-main");
+  const netenv = network === "9c-main" ? "main" : network;
 
-  let netenv;
-  if (network === "9c-main") {
-    netenv = "main";
-  } else {
-    netenv = network;
-  }
-
+  // FIXME: project version number hard coding: 1.
   const launcherDownloadUrl = getDownloadUrl(
     netenv,
     peerVersionNumber,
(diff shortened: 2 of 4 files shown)


src/main/update/launcher-update.ts
```diff
@@ -116,34 +107,40 @@ export async function update(update: Update, listeners: IUpdateOptions) {
...
+
+  // FIXME: project version number hard coding: 1.
+  const network = getConfig("Network", "9c-main");
+
+  let netenv;
+  if (network === "9c-main") {
+    netenv = "main";
+  } else {
+    netenv = network;
   }
```
BasixKOR, line 119:
```suggestion
  const netenv = network === "9c-main" ? "main" : network;
```

src/main/constants.ts
```diff
@@ -0,0 +1 @@
+export const DOWNLOAD_URL = "https://download.nine-chronicles.com";
```
BasixKOR, line 1: It may be better to put this alongside with updater source code, although I'm not so sure about this suggestion. Other constants are usually stored inside `config.ts`, but that doesn't look like a good place either.

src/main/update/launcher-update.ts
```diff
@@ -116,34 +107,40 @@ export async function update(update: Update, listeners: IUpdateOptions) {
...
+    peerVersionNumber,
+    "launcher",
+    1,
+    process.platform
+  );
+  const playerDownloadUrl = getDownloadUrl(
+    netenv,
+    peerVersionNumber,
+    "player",
+    1,
```
BasixKOR, line 132: It seems like figuring out the project version is out of scope for this PR. So take this as a reminder.

src/main/update/util.ts
```diff
@@ -0,0 +1,47 @@
...
+const FILENAME_MAP: { [k in NodeJS.Platform]: string | null } = {
+  aix: null,
+  android: null,
+  darwin: "mac.tar.gz",
+  freebsd: null,
+  linux: null,
+  openbsd: null,
+  sunos: null,
+  win32: "win.zip",
+  cygwin: null,
```
BasixKOR, line 45:
```suggestion
  cygwin: "win.zip",
```

Cygwin is a Windows environment, so this would work, but since Cygwin builds were deprecated a long time ago, I don't think this is a concern at all.

BasixKOR (approved):

src/main/update/launcher-update.ts
Atralupus: cool

src/main/constants.ts
Atralupus: I had hardcoded it because... I didn't have a good place to put it, but it's not all that bad to add `DownloadBaseURI` to `config.ts`

BasixKOR: Yeah if possible I think it makes sense to put in there.

src/main/update/launcher-update.ts
Atralupus: https://github.com/planetarium/9c-launcher/issues/1744
I'll update the body later

src/main/update/util.ts
Atralupus: ok 👍

Atralupus:
🚀

PR #1739, opened 2022-08-11
Add PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD to CI install

BasixKOR:
It should result in slightly improved CI speed.


BasixKOR:
Doesn't have a huge impact, but it is good enough to be included, I think.

longfin (approved):

PR #1741, opened 2022-08-11
Dedupe dependencies

BasixKOR:
This should help reducing the space and time of modules.

For reviewers: You don't need (or should) read the diff of the lockfile. It can be trivially reproduced by `yarn dedupe` and the effect would be minimal.


BasixKOR:
Although the Prettier bump wasn't my intention, I quite like it. It is much easier to read IMO.

moreal (approved):

PR #1742, opened 2022-08-11
Ignore errors on removing useless stores

BasixKOR:
Hopefully solves some of snapshot issues we have.

### DIFF
--- a/src/main/snapshot.ts
+++ b/src/main/snapshot.ts
@@ -296,7 +296,7 @@ export async function extractSnapshot(
 
 export async function removeUselessStore(
   blockchainStorePath: string
-): Promise<void[]> {
+): Promise<PromiseSettledResult<void>[]> {
   const listOfUselessDb = [
     "9c-main",
     "chain",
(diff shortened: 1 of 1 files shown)


src/main/snapshot.ts
```diff
@@ -379,7 +385,10 @@ export async function processSnapshot(
       mixpanel
     );
     snapshotPaths.push(stateSnapshotPath);
-    removeUselessStore(storePath);
+    await removeUselessStore(storePath).catch((error) => {
+      if (error.code === "ENOENT" || error.code === "ENOTDIR") return; // Ignore if not exist
+      console.error("Error while removing useless store: ", error);
+    });
```
longfin, line 391:
```suggestion
    try
    {
      await removeUselessStore(storePath);
    }
    catch (error)
    {
      if (error.code === "ENOENT" || error.code === "ENOTDIR") return; // Ignore if not exist
      console.error("Error while removing useless store: ", error);
    }
```

but it doesn't seem to a necessary work since it's a hotfix.

longfin (approved):

src/main/snapshot.ts
BasixKOR: While reviewing the comment I've noticed that `Promise.all` prevents other removals to be performed when a single one fails, so I've updated the PR accordingly.

longfin (approved):

PR #1777, opened 2022-08-19
[Fix] player path include network info

Atralupus:
Fix player path`${app.getPath("userData")}/player` -> `${app.getPath("userData")}/player/${netenv}`
and some refactoring...🙄


BasixKOR (approved):
It would've been nice to make a separate commit for refactoring stuff, but I guess this works too.

Atralupus:
> It would've been nice to make a separate commit for refactoring stuff, but I guess this works too.

Sorry In my defense... it's small refactoring

PR #1784, opened 2022-08-19
[Feature] Cleanup old player

Atralupus:
`player-update-temp` runs only the first time, So I delete old player in this script


BasixKOR (approved):
Seems reasonable to me.

Atralupus:
🚀

PR #1790, opened 2022-08-19
[Fix] File related bugs

Atralupus:
1. [I added network info in playerPath](https://github.com/planetarium/9c-launcher/pull/1777), but an error occurs because the recursive option doesn't have.
2. The modified launcher doesn't have `9c.exe`, but the delete player script didn't handle the error

### DIFF
--- a/src/main/update/player-update.ts
+++ b/src/main/update/player-update.ts
@@ -44,7 +44,7 @@ export async function playerUpdate(
   if (fs.existsSync(playerPath)) {
     fs.rmdirSync(playerPath, { recursive: true });
   } else {
-    fs.mkdirSync(playerPath);
+    fs.mkdirSync(playerPath, { recursive: true });
   }
 
   console.log("[player] Clean up exists player");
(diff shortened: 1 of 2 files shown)


src/main/update/player-update.ts
```diff
@@ -44,7 +44,7 @@ export async function playerUpdate(
   if (fs.existsSync(playerPath)) {
     fs.rmdirSync(playerPath, { recursive: true });
   } else {
-    fs.mkdirSync(playerPath);
+    fs.mkdirSync(playerPath, { recursive: true });
```
BasixKOR, line 47:
```suggestion
    await fs.promises.mkdir(playerPath, { recursive: true });
```

Synchronous functions block the thread while running, but I guess this is for other PRs.

BasixKOR (approved):

src/main/update/player-update.ts
Atralupus: @.@ I dont know why I did it
I'll fix it other PR Thanks 😵‍💫

PR #1792, opened 2022-08-19
[Fix] playerPath format

Atralupus:
using nodejs path...


Atralupus:
Sorry, Including also minor fix 👀

PR #1818, opened 2022-08-23
Fix tsc pre-commit hook command

BasixKOR:
This fixes the tsc command to not give each files as an argument, thus letting TypeScript check files based on the tsconfig.json.


Akamig (approved):

BasixKOR:
I'm skipping the CI because this PR doesn't introduce any changes to the application itself.

PR #1827, opened 2022-08-23
Removed `src/preload`

### LINKED ISSUE #1759: Remove `src/preload`
It appears to be almost unused.

`sentry.ts` is used, but it isn't used during the preload phase of the Electron (which `src/preload` implies) and therefore should be moved.

PrajwalBorkar:
Fixes #1759


BasixKOR (changes requested):
`sentry.ts` contains some code which is used in various places, you may want to move it instead

PrajwalBorkar:
Can you explain where I have to move the code?

BasixKOR:
Inside `src` should be fine (for now), although I think we should make some kind of utils folder someday

BasixKOR:
`initializeSentry` is also called from few places: notably `src/renderer/render.tsx`, `src/v2/render.tsx`, and `src/main/main.tsx`. I think this PR is okay to merge once we update those references!

BasixKOR (approved):

src/sentry.ts
```diff
@@ -1,26 +1,26 @@
...
-    init({
-      dsn: dsn,
-      release: version,
-    });
-    return;
-  }
-  console.debug("Sentry is disabled by config");
-}
+import { get, app } from "../config";
+import { version } from "../../package.json";
```
BasixKOR, line 2:
```suggestion
import { get, app } from "./config";
import { version } from "../package.json";
```

BasixKOR:
I'm closing this PR because the communication cost of this PR is too high to justify. The PR failed to mitigate the impact of the removal (which could be looked up by checking module references and imports), and I don't think stating the points to change one by one doesn't do any better than doing it myself.

PR #1832, opened 2022-08-24
[Chore] Restore e2e test

Atralupus:
Add "UseUpdate" option in `IConfig`

### DIFF
--- a/src/main/update/launcher-update.ts
+++ b/src/main/update/launcher-update.ts
@@ -69,6 +69,10 @@ export interface IUpdateOptions {
 }
 
 export async function update(update: Update, listeners: IUpdateOptions) {
+  if (!getConfig("UseUpdate", true)) {
+    return;
+  }
+
   const localVersionNumber: number = update.current ?? apvVersionNumber;
   const peerVersionNumber: number = update.newer;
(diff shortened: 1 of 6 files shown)


src/main/update/launcher-update.ts
```diff
@@ -69,6 +69,10 @@ export interface IUpdateOptions {
 }
 
 export async function update(update: Update, listeners: IUpdateOptions) {
+  if (!getConfig("UseUpdate", true)) {
```
BasixKOR, line 72:
```suggestion
  if (!getConfig("UseUpdate", process.env.NODE_ENV === "production")) {
```

Just a touch to disable updates during development.

src/main/update/launcher-update.ts
```diff
@@ -69,6 +69,10 @@ export interface IUpdateOptions {
...
+    return;
```
BasixKOR, line 73: Consider adding a log here.

Atralupus: Good

src/main/update/launcher-update.ts
Atralupus: Gooooood

BasixKOR (approved):

PR #1833, opened 2022-08-24
Fixed getMessage function

Akamig:
Fixed misbehaving getMessage function.


BasixKOR (approved):
Not very ideal, but I think it is good enough to be merged.

Specifically: I want the messages to be co-located with the states themselves so we can edit them more easily,

PR #1834, opened 2022-08-24
Fixed state on player-update-temp

Akamig:
Fixed states according to machine.


BasixKOR:
Doesn't this change effectively makes `player-update-temp` identical to the `player-update`?

Akamig:
We're going to refactor these player-launcher code after merged in development, so let's focus on keep it in working state

BasixKOR (approved):
Merging `player-update-temp` and `player-update` should be addressed in a subsequent PR.

PR #1864, opened 2022-08-30
[Fix] Rollback filename

Atralupus:
I have decided on the file name `win.zip`, `mac.tar.gz` but exists s3 file name is `Windows.zip`, `macOS.tar.gz`
We don't have a reason to change the file name, So I rollback file name
~~And fix default download base url~~


BasixKOR (approved):
Reasonable.

PR #1885, opened 2022-09-02
Fix IPC event name mismatch on updateMachine

BasixKOR:
https://github.com/planetarium/9c-launcher/blob/e85487b00a6c0aa9a26e330eedc1dbf33d94aa12/src/main/update/launcher-update.ts#L294

The event name doesn't match with the main IPC, causing the event to be ignored.


Atralupus (approved):
😞

PR #1927, opened 2022-09-21
Make direct state change while both player update and launcher update required

### LINKED ISSUE #1920: It causes blinking between player update and launcher update because of state machine goes to DONE when player update finished, and goes to launcher update directly after.
It causes blinking between player update and launcher update because of state machine goes to DONE when player update finished, and goes to launcher update directly after.
it seems fixable if we check IUpdate just before player update finishes and send state to launcher update.

_Originally posted by @Akamig in https://github.com/planetarium/9c-launcher/issues/1917#issuecomment-1249182308_

Akamig:
This Resolves #1920, but I'm not sure at certain points.

### DIFF
--- a/src/main/update/player-update.ts
+++ b/src/main/update/player-update.ts
@@ -1,4 +1,4 @@
-import { DownloadItem, app } from "electron";
+import { DownloadItem, app, ipcRenderer } from "electron";
 import { download, Options as ElectronDLOptions } from "electron-dl";
 import { IDownloadProgress } from "src/interfaces/ipc";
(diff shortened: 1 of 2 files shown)


BasixKOR (commented):
`invokeIpcEvent` should be invoked within the state listening to the event.

Akamig:
I haven't tested yet, but it seems to work I guess...?

Akamig:
Fixed condition, tested and confirmed it works like a charm.

BasixKOR (approved):
dunno, lgtm

Issue #1907, opened 2022-09-07
Support mac, linux updater


Atralupus:
Survey
- [x] Not working player updater in macOS, maybe permission issue, we should need to survey more
- [x] Not working open player in macOS...
- [ ] Not working launcher updater

Todo
- [x] Fix not found path error in player updater [mac, linux]

prev issue: https://github.com/planetarium/9c-launcher/issues/1900

sky1045:
https://github.com/planetarium/9c-launcher/blob/development/src/main/update/launcher-update.ts#L116

this condition always return true, so macos tries to update forever
because `executePath` will be `/Users/{username}/Library/Application Support/Nine Chonicles/player/main/9c`.
the executable file name should be `9c.app`, not `9c`

sky1045:
https://github.com/planetarium/9c-launcher/blob/development/src/main/main.ts#L369

for macos, node cannot execute `9c.app` directly. i made this work modifying the path to the executable file inside package.(`9c.app/Contents/MacOS/9c`)

i'm not sure why these happen after changing player path, so there would be better solution, but that's all i could solve

PR #1944, opened 2022-09-28
fix mac game path

sky1045:
i can run launcher on M1 Mac with this update, otherwise launcher tries to update player forever.
but i'm not sure how this can affect others.
let me know if something wrong can happen with this
https://github.com/planetarium/9c-launcher/issues/1907

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -188,7 +188,7 @@ export function getBlockChainStorePath(): string {
 
 export const REQUIRED_DISK_SPACE = 20n * 1000n * 1000n * 1000n;
 export const SNAPSHOT_SAVE_PATH = app.getPath("userData");
-export const MAC_GAME_PATH = path.join(playerPath, "9c");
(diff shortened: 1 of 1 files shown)


src/config.ts
```diff
@@ -188,7 +188,7 @@ export function getBlockChainStorePath(): string {
 
 export const REQUIRED_DISK_SPACE = 20n * 1000n * 1000n * 1000n;
 export const SNAPSHOT_SAVE_PATH = app.getPath("userData");
-export const MAC_GAME_PATH = path.join(playerPath, "9c");
+export const MAC_GAME_PATH = path.join(playerPath, "9c.app/Contents/MacOS/9c");
```
Atralupus, line 191: execute code is separated so this line will be fine

BasixKOR (approved):
Yeah, I think it is actually a regression since the path was specified as `9c.app/Contents/MacOS/9c` before.

Atralupus (approved):
Goood

PR #1955, opened 2022-10-19
Check node tip when selecting an RPC node

BasixKOR:
This PR proposes an alternative method to select an RPC node to connect using both the tip index and client counts. The configurator may specify the sampling rate to adjust how much client counts influence the node selection.

# New `config.json` field
## `RemoteClientSamplingRate: number`
The sampling rate to adjust the client count's influence on the node selection. The higher the number, the less impact the client count makes on the node selection.
If you want to completely disable this behaviour and use only the tip index to select the node, you can put `0`.

### DIFF
--- a/src/api.graphql
+++ b/src/api.graphql
@@ -363,6 +363,9 @@ mutation StageTxV2($encodedTx: String!) {
 query PreloadEnded {
   nodeStatus {
     preloadEnded
+    tip {
+      index
+    }
     appProtocolVersion {
       version
       extra
--- a/src/config.ts
+++ b/src/config.ts
@@ -57,7 +57,8 @@ export class NodeInfo {
   readonly graphqlPort: number;
   readonly rpcPort: number;
   readonly nodeNumber: number;
-  clientCount: number = 0;
+  clientCount = 0;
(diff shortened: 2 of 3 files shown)


area363:
https://app.circleci.com/pipelines/github/planetarium/9c-launcher/2340/workflows/1ab08c75-8a7b-4cdd-bd11-cb2d83413d04/jobs/10600
Is this a flaky test?

BasixKOR:
RPC selections are not parts of the test so probably yes.

Akamig:
> https://app.circleci.com/pipelines/github/planetarium/9c-launcher/2340/workflows/1ab08c75-8a7b-4cdd-bd11-cb2d83413d04/jobs/10600
> 
> Is this a flaky test?

Yup that e2e was quite flaky for a while.

area363 (approved):
LGTM Thank you for this feature! 🚀

longfin:
I think that removing stale nodes from candidates can be better than sorting with weight since stale nodes didn't tend to auto-recovery even reducing new connections via traffic control. cc @area363 @Akamig

PR #1966, opened 2022-10-28
Feat/collection reward diff

Akamig:
supersedes #1570

### DIFF
--- a/src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
+++ b/src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
@@ -303,13 +303,13 @@ export function MonsterCollectionContent({
               })}
             </ItemGroup>
             <ItemGroup key="system" title="System Rewards">
-              {systemRewards.map((item) => {
+              {systemRewards.map((item, index) => {
                 const sysRewardSuffix = item.name === "stage" ? "% DC" : "%";
-                const amount = item.amount[currentIndex ?? 0] + sysRewardSuffix;
-                const updatedAmount = item.amount[selectedIndex ?? 0] + sysRewardSuffix;
+                const amount = item.amount[currentIndex ?? 0];
+                const updatedAmount = item.amount[selectedIndex ?? 0];
                 return (
                   <Item
-                    key={item.name}
+                    key={index}
                     amount={isEditing ? "" : amount + sysRewardSuffix}
                     updatedAmount={
                       isEditing ? updatedAmount + sysRewardSuffix : ""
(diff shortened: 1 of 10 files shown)


src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
```diff
@@ -298,91 +294,33 @@ export function MonsterCollectionContent({
...
-                >
-                  <img src={apImg} height={48} />
-                </Item>
-              </ItemGroup>
-            )}
+            <ItemGroup key="system" title="System Rewards">
+              {systemRewards.map((item) => {
+                const amount = item.amount[currentIndex ?? 0];
+                const updatedAmount = item.amount[selectedIndex ?? 0];
+                const sysRewardSuffix = item.name === "stage" ? "% DC" : "%";
```
BasixKOR, line 309:
```suggestion
                const sysRewardSuffix = item.name === "stage" ? "% DC" : "%";
                const amount = item.amount[currentIndex ?? 0] + sysRewardSuffix;
                const updatedAmount = item.amount[selectedIndex ?? 0] + sysRewardSuffix;
```

This way, we don't need to concatenate the suffix every time we display them.

src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
```diff
@@ -298,91 +294,33 @@ export function MonsterCollectionContent({
...
+                return (
+                  <Item
+                    key={item.name}
```
BasixKOR, line 312: As the index will remain stable for the lifetime of the session, I think it's safe to use `index`, although we don't really need to change the code to use the indices.

src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
```diff
@@ -298,91 +294,33 @@ export function MonsterCollectionContent({
                     title={itemMeta.name}
                     isUpgrade={selectedAmount?.gte(itemAmount)}
                     updatedAmount={selectedAmount?.toString()}
+                    isDiff={true}
```
BasixKOR, line 298:
```suggestion
                    isDiff
```

src/v2/views/MonsterCollectionOverlay/MonsterCollectionContent.tsx
```diff
@@ -203,6 +203,7 @@ export function MonsterCollectionContent({
                   value={amount}
                   onChange={(e) => setAmount(e.target.value)}
                   maxLength={6}
+                  pattern={"[0-9]*"}
```
BasixKOR, line 206: Why not `type="number"`?

BasixKOR (approved):

PR #1972, opened 2022-11-01
Apply Relative tip comparison and ClientCount-weighted RPC Node Selector

Akamig:
This PR Adds #1973 

- Relative tip comparison between available nodes, 
if node's tip is falling behind under threshold (default: 20 block), it drops from node list.
- ClientCount-weighted random RPC node selection
when selecting node, instead of unconditionally selecting least used node, Use clientCount based weight (more client = less weight) and randomly select based on that. from this, We can stay between "Almost Deterministic Node Selection (but when best node is not actually working it stucks)" and "Purely Random Selection(which possibly select the worst node)".

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -129,11 +129,9 @@ const NonStaleNodeList = (
 ): NodeInfo[] => {
   if (staleThreshold < 0) {
     return nodeList;
-  } else {
-    return nodeList
-      .sort((a, b) => b.tip - a.tip)
-      .filter((node) => node.tip >= nodeList[0].tip - staleThreshold);
   }
+  const maxTip = Math.max(...nodeList.map((node) => node.tip));
+  return nodeList.filter((node) => node.tip >= maxTip - staleThreshold);
 };
 
 const clientWeightedSelector = (nodeList: NodeInfo[]): NodeInfo => {
--- a/src/config.ts
+++ b/src/config.ts
@@ -127,9 +127,13 @@ const NonStaleNodeList = (
   nodeList: NodeInfo[],
   staleThreshold: number
 ): NodeInfo[] => {
-  return nodeList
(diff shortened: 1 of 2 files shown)


Akamig:
앗 필터에서 실수했ㄷ

src/config.ts
```diff
@@ -123,6 +123,31 @@ const NodeList = async (): Promise<NodeInfo[]> => {
...
+const NonStaleNodeList = async (
+  nodeList: NodeInfo[],
+  staleThreshold: number
+): Promise<NodeInfo[]> => {
+  return nodeList
+    .sort((a, b) => b.tip - a.tip)
+    .filter((node) => node.tip <= nodeList[0].tip - staleThreshold);
+};
+
+const clientWeightedSelector = async (
```
BasixKOR, line 135:
```suggestion
const clientWeightedSelector = (
```

Doesn't need to be async.

src/config.ts
```diff
@@ -123,6 +123,31 @@ const NodeList = async (): Promise<NodeInfo[]> => {
   return nodeList;
 };
 
+const NonStaleNodeList = async (
```
BasixKOR, line 126:
```suggestion
const NonStaleNodeList = (
```

src/config.ts
```diff
@@ -229,18 +254,13 @@ export const installerUrl = path.join(DEFAULT_DOWNLOAD_BASE_URL, installerName);
 
 export async function initializeNode(): Promise<NodeInfo> {
   console.log("config initialize called");
-  const nodeList = await NodeList();
+  const relativeTipLimit = get("RemoteClientStaleTipLimit", 20) || Infinity;
```
BasixKOR, line 257: Is the usage of `Infinity` intended?

BasixKOR (approved):

BasixKOR (approved):

PR #2008, opened 2022-11-25
Use "Activation Code" instead of "Invitation Code"

Akamig:


src/v2/views/RegisterView/index.tsx
```diff
@@ -64,6 +64,7 @@ function RegisterView() {
         address={key?.address}
         onSubmit={onSubmit}
         useActivitionKey
+        useLink={"https://nine-chronicles.com/start?step=2"}
```
longfin, line 67:
```suggestion
        urlForActivationCode={"https://nine-chronicles.com/start?step=2"}
```
it seems that `use`-prefix  has been intended for `bool` (e.g., `useActivation`)

longfin, line 67: I also guess that current URL(`"https://nine-chronicles.com/start?step=2"`) might be changed for various reasons. (In particular, query strings such as `step=2` often change according to the specifications of the service.) in that case, we need to update the whole binary... 😢 

- How about using config instead hard coded value?
- Or, leveraging indirection URL seems an option too.

cc @Gilhwan @rareboe @minmboy

Akamig: I just removed this specific option entirely from this component because it's usage seems too narrow. instead I moved it to RegisterView itself.

longfin (approved):

PR #2013, opened 2022-11-29
[Fix] update required rule

Atralupus:
before the test


Akamig (approved):

Atralupus:
we need to unit test... 😭

PR #2014, opened 2022-11-29
Fix claim stake reward not working.

Akamig:


moreal (commented):
Maybe https://github.com/planetarium/9c-launcher/blob/d078a131fe378827dd00f1186a8b4b951de3ca14/src/main/headless/headless.ts#L106-L121 also should be fixed.

PR #2033, opened 2022-12-12
Deprecate Keystore related Headless API

Akamig:


moreal (commented):
It seems there are usages of APIs deleted in this pull request. And they're making the `styles` job fail.

moreal (approved):

Akamig:
Now it's ready to review, at least from what I look. 
a1830022466_10

Please Ignore Build Failure from not edited V1 codes.
AAAAAAAAAAAA YES

__tests__/main/headless/validation.spec.ts
```diff
@@ -1,21 +1,19 @@
 import { describe, it, assert } from "vitest";
-import { Validation } from "../../../src/main/headless/validation";
-import { HEADLESS_PATH } from "../../constant";
+import AccountStore from "src/stores/account";
```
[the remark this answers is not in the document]
Akamig: In long term plain I'm trying to salvage some usable tests from current tests then adding new tests, so for now it stays.

Akamig:
Issue of current account.ts (accountStore) code in my view
1. Whether 'address' have 0x prefix or not is **obscure**, whether in V3 level, keyring PPK level, or as given argument. so we should decide 0x should be always included or not and force it.
2. Current methods (or actions in MobX term) of accountStore, especially existing private key related logic is partially depending on UI flow.
    - (for example, RegisterView setting privatekey is purely depending on expected behavior of LobbyView executing `getPrivateKeyAndForget`, which immediately use privatekey and discard.),
3. So 'unexpected' UI flow may trigger exception or unintended behavior.
    - (From the former example, as we get private key and **immidiately forget** on first game client launch, exit game client and try to relaunch will fail as there's no raw privatekey.)
4. Even it's momentarily, we still storing and passing rawPrivateKey from accountStore, which is security vulnerability.
5. I have no confidence on synchronization of `account` and `address` in accountStore, and whether `address` always exists in keyring entry.
6. duplicated key processing in keyring is quite an issue.

src/stores/account.ts
```diff
@@ -61,14 +98,117 @@ export default class AccountStore implements IAccountStore {
...
+
+  @action
+  isValidPrivateKey = (privateKey: string): boolean => {
+    try {
+      createAccount(privateKey);
+    } catch (e) {
+      return false;
     }
+    return true;
   };
```
moreal, line 179: `isValidPrivateKey` function seems able to be separated as a non-class function.

src/stores/account.ts
```diff
@@ -61,14 +98,117 @@ export default class AccountStore implements IAccountStore {
...
 
   @action
-  changeAddress = (index: number) => {
-    if (index >= 0 && index < this.addresses.length) {
-      this.selectedAddress = this.addresses[index];
+  setActivationKey = (activationKey: string) => {
+    this.activationKey = activationKey;
+  };
```
moreal, line 103: > Question need to be answered : Is accountStore is too huge? we're using store as some kind of singleton instance, is too many functions focused on store okay?

In the above context, should `setActivationKey` be separated as another store?

src/v2/views/RevokeView.tsx
```diff
@@ -1,18 +1,19 @@
...
+import { Address } from "src/interfaces/keystore";
 import Button from "../components/ui/Button";
-import { ipcRenderer } from "electron";
 import { useHistory } from "react-router";
 
 const transifexTags = "v2/revoke-view";
 
 function RevokeView() {
   const account = useStore("account");
+  const [address, setAddress] = useState<Address>(account.keyring[0].address);
```
moreal, line 16: It assumes users will come to `RevokeView` when only they have an address at leat, right?

src/stores/account.ts
Akamig: Seperated By adding `isValidPrivateKey` in @planetarium/account-localm

src/stores/account.ts
Akamig: Is a bit complicated but we need ActivationKey in usually accountStore-called environment. so I'm going to leave it now, but your point also valid.

Akamig (commented):
There was no reviewer so I reviewed thrice and found no critical issue. so going to merge it.

PR #2041, opened 2022-12-16
Deprecate Action-Tx Related Headless Feature

Akamig:
This resolves #2040, and must not be merged before #2033 merged.

PR #2030, opened 2022-12-09
Replaced APV Related headless feature

Akamig:
queryApv -> Replaced with GraphQL Query
verifyApv -> implemented with `@noble/secp256k1` 
analyzeApv -> Extra flattener implemented

Issue #1713, opened 2022-08-02
Sunsetting the legacy interface


BasixKOR:
This issue will be addressed after complete legacy monster collection sunsetting

- [x] Remove `PreferLegacyInterface` configuration
- [x] Move `src/v2` into `src/renderer`
  - [x] Merge some symlinks (Notably src/v2/i18n and src/v2/stores)
  - [x] Remove the legacy interface entirely
- [x] Move `src/main/v2` into `src/main`

PR #2049, opened 2022-12-26
Detach NineChronicles.Headless from Launcher.

### LINKED ISSUE #1913: Detach Headless from Launcher
Currently, we depends on the headless for these:
- [x] Running the node
- [x] #1914
- [x] #2017 
- [x] Signing the transaction

Although the first one cannot be done inside the launcher, we can implement the rest in the launcher. This issue serves as a tracking issue for decoupling the headless from the launcher.

Akamig:
I'm so honored to be the one announce that we're finally detaching headless submodule from launcher,
And sunsetting V1 launcher code which was diligently developed by many people ahead of me.
Many thanks to @BasixKOR who tremendously helped my understandings on this project.

**This Pull Requests Resolves #1913.**

# Changelog
  ## Replace features provided by NineChronicles.Headless in 9c-launcher.
Replace headless store & features used in renderer, mainly located in `src/main/headless`

  ### APV 
#2030 includes all APV related headless feature replacement, replaced code mainly located in `src/utils/apv`
  - `query` 
      - This was only been used in update related logic, so this was directly implemented in `src/main/update/check/` as `getNodeApv`.
  - `analyze`
      - I needed to analyze two types of APV, GQL query result which already provides separated properties in single APV string, and **slash-delimited, APV token string** from config.json, which needed to be dissected and decode. so I implemented into separate functions to output same type of APV object. (IApv)
      -  also in the process of analyzing APV, I needed specific bencodex value flattening logic in order to objectify `extra` value which could be any bencodex value, so that logic is implemented in `src/utils/bencodex`.
  - `verify` 
      - This simply take APV Object, and public key (usually value of TrustedAPVSigner), Reorganize object properties into single Buffer in right encoding and get SHA-256 hash, fetch signature, and verify secp256k1 with provided SHA-256 hash, signature, and public key.
---
 ### Keystore
PR #2033 includes all keystore related headless feature replacement, replaced code mainly located in `src/stores/account`
 #### `account.ts`
 #### General Behavior
I supercharged AccountStore with `planetarium/sphere`, so It basically works like this.
  - in the initialization time, accountStore fetches all V3 keystore files from default planetarium keystore paths, (provided by sphere) and store them in keyring, as `ProtectedPrivateKey` type
  - `ProtectedPrivateKey` object stores UUID of V3 file, address, and path to that file.
  - From the keyring, I can select a PPK and make `Account` instance of sphere with passphrase and set that instance in accountStore, which can be reused for signing transactons throughout lifecycle.
      - `address` value of accountStore is setted when generating Account instance (`getAccount`), and importing raw PrivateKey from outside, (`importRaw`), which makes easier to get currently 'selected', 'Setted Accounts address`. it seems little bit finicky since there's no limit on setting address and cause 'de-synced' from current account / raw privatekey if misused, but from usage perspective, I couldn't find any issue. so there we have.
#### Raw Private Key Issue
  - **We sadly still have raw private key handling logics in parallel of Account instance in accountStore.** 
      - For security measures, I store it in `Buffer`, which is dynamically reassignable real memory space, and also implemented only getter of privatekey, `getPrivateKeyAndForget`. Promise that literally wipe out memory space with zeros immediately after raw privatekey return. effectively making all raw privatekey one-time use only. with this method, we truly wipe privatekey data on-time instead of depend on garbage collection cycle, possibly leave value in memory even after discard.
      - It's mainly used for Register, Import, ForgotPassword and Recover, but most of these use cases are usually trigger `getPrivateKeyAndForget` right after `setPrivatekey`, only momentarily used for inter-page value store.
      - The most problematic usage of this is that we must provide "raw privatekey" as argument on starting game client, which is big vulnerability, and should be replaced with more secure method of ipc and different approach on authentication.
      - With the former issue, since 'Client gets raw privatekey' and we "IMMEDIATELY FORGETS IT", we can't relaunch game properly without relaunch the launcher and login again, specifically, we "CAN" restart the game client, but it will run with 'wiped' raw privatekey of "00000000000000000000000000000000000000000000000000000000000000', which is, surprisingly a valid private key.
#### Renderer View
  - Other than that, I successfully replaced ipc Call to standalone CLI `key` commands with the help of sphere and accountStore in various views.
---
 ### Action-Tx
- PR #2041 includes all Action-Tx related headless feature replacement.
- In the beginning phase, I checked which actions are used on where, and found out 'ActivateAccount' should be implemented in graphql endpoint, so @U-lis did it on headless https://github.com/planetarium/NineChronicles.Headless/pull/1765 and I updated `schema.graphql` accordingly, thanks.
- With that, all necessary `actionTxQuery` queries were added into `api.graphql`.
- Because I replaced most of queries with [Apollo useQuery hooks](https://www.apollographql.com/docs/react/api/react/hooks/) former `useTx` hook which strongly depended on Headless and steps of 'Action -> Create Transaction -> Sign -> Stage` needed to be heavily modified, originally tried to integrate Apollo hooks into useTx so we can process error validation collectively, but failed.
- So currently Error handling of initial `actionTxQuery` through Apollo is managed case-by-case, `useTx` get TX string from former query, **and it only manages signing and staging and it's error.**
#### Deprecation of Legacy Monster Collection
- On the other hand, we decided to not implement legacy monster collection related query which was not in use already, resulting deprecation of `src/collection` (V1 MC)
#### Integration of Transfer into Launcher.
- Until now, 'Send NCG' was ran as separated process, which doesn't share any state or memory with launcher and has it's own state store. it has it's advantages of being independent module, but I found difficulties in sharing accountStore's state and it's value on initialization of `transfer`. 
- In order to solve that I managed to make it from window to Overlay view, just like Settings and V2 MonsterCollection. from this. I was able to call useStore() to use contents of accountStore and sign NCG transfer transactions.
- Also, in order to port previously used code, I introduced new `src/store/transfer` store to manage functions used in `TransferAssetOverlay`
- Original `src/transfer` uses `graphql-requests` to send queries, which is different from main launcher logic using Apollo, it needs GraphQLClient and SDK from `getSdk()` in order to send any queries. in short, it needs GQL endpoint address of main process. it was not easy to put it in constructor of Store to do that, I put `useStore()` and ipc invoke of `get-node-info` on `src/renderer/App.tsx`, so It can be configured just after end of main process initialization and render of any page. welcome any suggestion for better place to put it.
---

## Byebye, V1 Launcher!
PR #2048 includes all V1 code removal and some headless-related code cleaning in the process.

### Why?
- Because V1 launcher uses tons of deprecated gql queries and headless dependency, It seems not viable enough to fix considering lack of user pool. also we just deprecated V1 monster collection, so Instead of replace and remove logics I decided to remove V1 from current codebase and make V2 just, launcher.
#### V1 Deprecation
- Mostly followed #1713 
- Remove most of `src/renderer` including components, views, providers.
- Move contents of V2 into `src/renderer`
- Merge all `utils` folder into single `src/utils`
- Rename v2.html to index.html
- Fix GraphQL codegen to generate code in renderer
- Remove V2 folder
- Adjust import paths of each modules
- Change webpack setting
#### Headless Cleaning
- Remove all headless/standalone usage from `src/main/main`
- Remove all headless related scripts in circleci.yml, package.json, and `src/scripts`
- Remove all except `remoteHeadless.ts` from `src/main/headless`
- Since we no longer need to load snapshots of chain data, removed `src/preload` (resolving #1759) and all snapshot related code.
- Since there's no cache to clear, remove "Clear Cache" button from settings
- Properties only for headless in `config.json` removed from interface.
#### ETC
- Remove mining related features (since we're deploying PBFT)
- Modify Preload State Machine
- Remove stale tests using headless

**AND FINALLY, REMOVING `NineChronicles.Headless` SUBMODULE FROM THIS REPOSITORY**

Et c'est fini.


Akamig:
Done.
Review Ready.
Fully Detached.
Headless-less.

Akamig:
마쳐봅시다.

PR #2057, opened 2023-01-11
Fix max length of deposit

jckdotim:
Deposit can be over 10M+ NCG


Akamig:
This 6 digit limit was temporary fix that UI breaking issue currently have with over 6-digit-numbers,

So we need to change not only the limit but the **shape of Input field** too. I'm trying some alternative UIs and will be fixed on that PR. and I promise it will be before the next release. so I'm closing this issue for now, but ultimately I'm targeting displaying figures over 9 digits correctly.

Akamig:
This 6 digit limit was temporary fix that UI breaking issue currently have with over 6-digit-numbers,

So we need to change not only the limit but the **shape of Input field** too. I'm trying some alternative UIs and will be fixed on that PR. and I promise it will be before the next release. so I'm closing this issue for now, but ultimately I'm targeting displaying figures over 9 digits correctly.

PR #2063, opened 2023-01-12
Fix False Timeout on Send NCG

### LINKED ISSUE #2056: Send NCG execution always considered as timed out even if it succeed

Akamig:
this resolves #2056


Akamig:
THANK GOD

PR #2064, opened 2023-01-12
[Feature] Remove usePreload

Atralupus:
I created `useIsHeadlessAvailable` hooks instead of `usePreload` hooks


src/stores/game.ts
```diff
@@ -17,6 +17,8 @@ export default class GameStore {
   private _port: number | undefined;
 
   public constructor() {
+    makeObservable(this);
```
Atralupus, line 20: https://mobx.js.org/enabling-decorators.html

Akamig: Tested and working.

Issue #2051, opened 2022-12-29
Problem of changing apv for no reason


Atralupus:
I am experiencing an issue with the APV in `main.log` on my computer
```
[2022-12-28 15:19:26.247] [error] An error has occurred while checking updates Error: Command failed: .\NineChronicles.Headless.Executable apv analyze --json 100350/ZHU4OmxhdW5jaGVydTQyOjEvYzExOGQ2Njg1MWQyYzI2MWY0YjNkMmIzMzMxODE0ODFkMGZmOWVkOHU2OnBsYXllcnU0MjoyLzhlMjczNWIwNjY5NzViOGJlYzQwYTFkOGNjMGU4MzUyODBjZDZjZmV1OTp0aW1lc3RhbXB1MTA6MjAyMi0xMi0yN2U=
Error: Not a valid app protocol version token.  Failed to find the second field delimiter.
```
Occurred `Not a valid app protocol version token` error

The APV listed in the error message is different from the original APV of
```
100350/6ec8E598962F1f475504F82fD5bF3410eAE58B9B/MEQCIBggvxpQiG5Ni5Eu2kCI2eWpyTiNMYEe+DRki3nMjVY1AiAS8faVt5RiBHplLTfwvZjy2+vMsHB29jfDhZPWiSQOOg==/ZHU4OmxhdW5jaGVydTQyOjEvYzExOGQ2Njg1MWQyYzI2MWY0YjNkMmIzMzMxODE0ODFkMGZmOWVkOHU2OnBsYXllcnU0MjoyLzhlMjczNWIwNjY5NzViOGJlYzQwYTFkOGNjMGU4MzUyODBjZDZjZmV1OTp0aW1lc3RhbXB1MTA6MjAyMi0xMi0yN2U=
```
It seems that the `6ec8E598962F1f475504F82fD5bF3410eAE58B9B/MEQCIBggvxpQiG5Ni5Eu2kCI2eWpyTiNMYEe+DRki3nMjVY1AiAS8faVt5RiBHplLTfwvZjy2+vMsHB29jfDhZPWiSQOOg==` portion of the APV has disappeared

I am not sure if this is a bug or not 🤔

Atralupus:
Before, we had launched 100350-2(only player update)

PR #2074, opened 2023-01-20
[Fix] Adjust `encounter different version` event APV values

Atralupus:
- Fix `NodeStatus` query
- Fix `encodeTokenFromHex` util


Akamig (approved):

Akamig:
this resolves #2051

PR #2081, opened 2023-01-29
Fix Keystore Duplication - Improper Key Handling while forgot password process

Akamig:


maxswjeon:
Seems `src/utils/useActivation.ts` isn't complying with the prettier format.

PR #2091, opened 2023-02-02
Fix Base64 I/O accordingly to Libplanet

### LINKED ISSUE #2090: APV Encode/Decode Logic output differs FromToken/FromGQL
APVs are different, ignore. (local: 100360/AB2da648b9154F2cCcAFBD85e0Bc3d51f97330Fc/MEUCIQDNbUbgTvf3ksHLGIu83XraKs4rNl/JTxeyCCnKkVQ1QgIgfDae3/O/B39arqF/E/PAp601/UmPi4ML5wUPBxaDKbU=/ZHU4OmxhdW5jaGVydTQyOjEvY2RkZTE5ZjEyMzUxNWIyMWViOWQ5OTYwM2ZhOWRmNzM3ZTE3MmJjZnU2OnBsYXllcnU0MjoxLzhmZjdkNWEwYmI4YmM1MjNjZDM4NzVjNGY5NTlkNGRhMTNlYzkwY2F1OTp0aW1lc3RhbXB1MTA6MjAyMy0wMi0wMWU=,
remote: 100360/AB2da648b9154F2cCcAFBD85e0Bc3d51f97330Fc/MEUCIQDNbUbgTvf3ksHLGIu83XraKs4rNl.JTxeyCCnKkVQ1QgIgfDae3.O.B39arqF.E.PAp601.UmPi4ML5wUPBxaDKbU=/ZHU4OmxhdW5jaGVydTQyOjEvY2RkZTE5ZjEyMzUxNWIyMWViOWQ5OTYwM2ZhOWRmNzM3ZTE3MmJjZnU2OnBsYXllcnU0MjoxLzhmZjdkNWEwYmI4YmM1MjNjZDM4NzVjNGY5NTlkNGRhMTNlYzkwY2F1OTp0aW1lc3RhbXB1MTA6MjAyMy0wMi0wMWU=)

[2023-02-02 14:13:49.937] [error] An error has occurred while checking updates Error: Invalid Bencodex Value.

Akamig:
https://github.com/planetarium/libplanet/blob/e9ba50938e53aaa87fcf3d3ec943201b191b21e0/Libplanet.Net/AppProtocolVersion.cs#L97

We must replace "/" with "." due to delimiter of APV is overlapping as "/".
this resolves #2090.


src/utils/apv.ts
```diff
@@ -17,7 +17,7 @@ export function analyzeApv(token: string | AppProtocolVersionType): IApv {
 
 function analyzeApvFromToken(token: string): IApv {
   const [version, signer, signature, rawExtra] = token
-    .replace(".", "")
+    .replaceAll(".", "")
```
moreal, line 20: APV seems to use the `'.'` character as a replacement for the `'/'` character.

So isn't it dangerous to remove all `'.'` characters?

For example, when `rawExtra` is `"...."` (means `"////"` in base64 standard), it will be `""`.

src/utils/apv.ts
```diff
@@ -27,7 +25,9 @@ function analyzeApvFromToken(token: string): IApv {
     raw: token,
     version: parseInt(version),
     signer: signer.replace("0x", ""),
-    signature: Buffer.from(signature, "base64").toString("hex"),
+    signature: Buffer.from(signature.replaceAll(".", "/"), "base64").toString(
```
moreal, line 28: Though I cannot reference the above code not changed in this pull request, the `const extra = decode(Buffer.from(rawExtra, "base64"))!;` part also should be fixed by replacing `'.'` with `'/'` like signature does.

longfin: [It seems so](https://github.com/planetarium/libplanet/blob/e9ba50938e53aaa87fcf3d3ec943201b191b21e0/Libplanet.Net/AppProtocolVersion.cs#L97-L101). thanks for pointing out.

src/utils/apv.ts
Akamig: Get safety confirm from @swen

longfin (approved):

moreal (approved):

PR #2110, opened 2023-03-06
feat: Activation Flow Update

maxswjeon:
This updates activation flow

- [x] Update Activation Flow
    - [x] Split UI steps
        - [x] Change RegisterView to allow subviews
        - [x] CreateKeySubview
        - [x] ActivationKeySubview
        - [x] ActivationWaitSubview
        - [x] ActivationResultSubview
    - [x] Split `useActivation()` hook
        - [x] `useActivate()` hook
        - [x] `useActivationStatus()` hook


maxswjeon:
This PR is not tested. Please do not commit before testing

src/renderer/views/RegisterView/subviews/ActivationWaitSubview.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+interface Props {
+  setState: (state: RegisterState) => void;
+}
+
+function ActivationWaitSubview({ setState }: Props) {
+  const activate = useActivate();
+  const { activated, error } = useActivationStatus();
+
+  useEffect(() => void activate(), [activate]);
```
Akamig, line 17: Need error handling according to error types of `useActivate()`, 
Recommend define and export error types in `src/interfaces` for better modularity.

also we would need state definition on failure of this hook.

src/renderer/views/RegisterView/subviews/ActivationResultSubview.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <p>Note: We strongly recommend you to backup your keystore.</p>
+      <Button
+        layout
+        variant="primary"
+        centered
+        css={{ width: 200 }}
+        onClick={() => setState("enterActivationCode")}
+      >
+        Activate
+      </Button>
```
maxswjeon, line 33: - [ ] Change button text
- [ ] Change redirection

src/renderer/views/RegisterView/subviews/ActivationResultSubview.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <p>Whoa it failed</p>
+      <Button
+        layout
+        variant="primary"
+        centered
+        css={{ width: 200 }}
+        onClick={() => setState("enterActivationCode")}
+      >
+        Activate
+      </Button>
```
maxswjeon, line 52: - [ ] Change button text
- [ ] Where should we go when activation fails?

src/renderer/views/RegisterView/subviews/ActivationWaitSubview.tsx
maxswjeon: - [x] Handle error from `useActivate()`
- [ ] Move error types to `src/interfaces`

src/renderer/views/RegisterView/subviews/ActivationResultSubview.tsx
```diff
@@ -0,0 +1,68 @@
...
+    </>
+  );
+}
+
+function ActivationFailSubview({ setState }: ResultProps) {
+  return (
+    <>
+      <H1>Activation has failed</H1>
+      {/* TODO: Add description here */}
+      <p>Whoa it failed</p>
```
maxswjeon, line 46: Please suggest appropriate text here  @Gilhwan

src/config.ts
```diff
@@ -234,6 +234,9 @@ export const DEFAULT_DOWNLOAD_BASE_URL = "https://release.nine-chronicles.com";
 export const PLAYER_METAFILE_VERSION = 2;
 export const installerName = "NineChroniclesInstaller.exe";
 
+// TODO: Update activation documentation link
+export const ACTIVATION_DOCUMENTATION_LINK = "";
+
```
maxswjeon, line 239: Please update this link when documentation is ready @Gilhwan

src/renderer/views/RegisterView/subviews/ActivationKeySubview.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <>
+      <H1>Activate your address</H1>
+      <p style={{ marginBlockEnd: 54 }}>
+        You need an activation code to activate your Nine Chronicles address. If
+        you already have one, you can paste it below and activate it now.
+      </p>
+      <ExtLink target="_blank" href="">
+        Get the code
+      </ExtLink>
```
maxswjeon, line 35: Link for `Get the code` is missing

src/utils/useActivate.ts
```diff
@@ -0,0 +1,117 @@
...
+
+type ActivationSuccessResult = {
+  result: true;
+  txId: string;
+};
+
+type ActivationFailResult = {
+  result: false;
+  error: {
+    error: unknown;
```
longfin, line 23: Could we specify a more specified type instead of `unknown`?

src/renderer/views/RegisterView/subviews/ActivationKeySubview.tsx
longfin: also, It may be useful if the address is configurable.

longfin (approved):
I'll continue to review after the remaining work on here has been finished. 🙏

src/renderer/views/RegisterView/subviews/ActivationKeySubview.tsx
Akamig: Previously I made ActivationCodeLink entry in config.json to get this link as I remember, I guess we can use same method as before.

src/utils/useActivate.ts
maxswjeon: The error is not thrown by internal code, so it's hard to determine the type of the error.
Also, the error instance is the same with the error instance that has been catched, so it's original type is `unknown` too.

src/renderer/views/RegisterView/subviews/ActivationKeySubview.tsx
maxswjeon: Made the link to read from the config file.

maxswjeon:
ExtLink is not opening on new window

src/utils/useActivate.ts
maxswjeon: @longfin Is this a major issue that blocks merge? If it isn't, can I consider it resolved and continue?

src/renderer/views/RegisterView/subviews/ActivationResultSubview.tsx
maxswjeon: Resolved in latter commit.

src/utils/useActivationStatus.ts
```diff
@@ -0,0 +1,33 @@
...
+  activated: boolean;
+}
+
+/**
+ * Queries the activation status of the current account.
+ *
+ * @returns {ActivationResult} A object with two properties: `loading` and `activated`. They are pretty self-explanatory.
+ */
+
+export function useActivationStatus(usePolling?: boolean): ActivationResult {
```
Akamig, line 17: I'd prefer assigning default value for this.

src/renderer/views/MissingActivationView.tsx
```diff
@@ -35,10 +35,10 @@ function MissingActivationView() {
   const history = useHistory();
   const account = useStore("account");
 
-  const { activated } = useActivation(false);
+  const { activated } = useActivationStatus();
```
Akamig, line 38: Isn't `useActivationStatus()` requires `usePolling` parameter?
You haven't specified on other `useActivationStatus()` usage too.

src/renderer/components/ActivationKeyForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+  onSubmit: (data: FormData) => void;
+}
+
+export interface FormData {
+  activationKey: string;
+}
+
+const ACTIVATION_KEY_REGEX = /^[0-9a-f]+\/[0-9a-f]{40}$/;
+
+type ActivationStatus =
```
Akamig, line 20: Making this type as enum rather than string literal union would be likely preferred, as we can separate message itself from state value itself, making editing message more reliable.

src/renderer/components/ui/Carousel.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+type Props = {
+  children?: JSX.Element[];
+  page: number;
+};
+
+function Carousel({ children }: Props) {
```
Akamig, line 8: Can you more elaborate on this component? is this stub code?

Akamig (commented):
Some.... questions?

src/utils/useActivate.ts
```diff
@@ -0,0 +1,102 @@
...
+            step,
+          },
+        };
+      }
+
+      return {
+        result: true,
+        txId: txData.stageTransaction,
+      };
+    } catch (e: unknown) {
```
Akamig, line 90: If you're using unknown as type of Error, I hope there's type assertion in catch such as
```
  if (error instanceof Error) {
    errorMessage = error.message;
  }
```

src/utils/useActivationStatus.ts
maxswjeon: Since the type of `usePolling` is `boolean | undefined`, the default value is `undefined` when not specified. Also, since `undefined` is a falsy value, it works as a default `false` value.

src/renderer/views/MissingActivationView.tsx
maxswjeon: `usePolling` parameter is optional, not an requirement.

src/renderer/components/ui/Carousel.tsx
maxswjeon: Oh, this is a component that was intended to be used in `ActivationWaitSubview`, but the content was not supplied. I'll remove it since it is more like a stub code.

src/utils/useActivationStatus.ts
Akamig: if then, explicitly specify default behavior like `(usePolling: boolean = false)` would be more strict rather than rely on falsy-ness. small difference but still.

maxswjeon: Resolved in [7e1577d](https://github.com/planetarium/9c-launcher/pull/2110/commits/7e1577d56067ba324483d7b3d5591b85ae10720c)

src/renderer/components/ActivationKeyForm.tsx
maxswjeon: I'm a little bit against using enums in typescript because of this [post](https://dev.to/dvddpl/whats-the-problem-with-typescript-enums-2okj), and the message is not used in this case.

src/utils/useActivate.ts
maxswjeon: I used that to propagate error to the parent without modification.

I know that I can assert the type but what we should do if the assertion fail?
```typescript
catch (error: unknown) {
    if (error instance of Error) {
        return {
            result: false,
            error: {
                error,
                step
            }
        }
    }
    // What should we do here?
    // Doing nothing results the function return signiture to be ActivationResult | undefined
}
```

src/renderer/components/ui/Carousel.tsx
maxswjeon: Resolved in [053492c](https://github.com/planetarium/9c-launcher/pull/2110/commits/053492c781882a9d4ec624d01190fcc635d5187d)

src/utils/useActivate.ts
maxswjeon: Maybe...?
```typescript
catch (error: unknown) {
    if (error instance of Error) {
        return {
            result: false,
            error: {
                error,
                step
            }
        }
    }
    return {
        result: false,
        error: {
            error: new Error("Unknown error occurred"),
            step
        }
    }
}
```

Akamig: I guess just re-throwing might be simple answer for this unusual case?
https://stackoverflow.com/a/67160588

maxswjeon: It seems a good way, applied on [053492c](https://github.com/planetarium/9c-launcher/pull/2110/commits/053492c781882a9d4ec624d01190fcc635d5187d)

src/utils/useActivate.ts
maxswjeon: Fixed on [3bbe211](https://github.com/planetarium/9c-launcher/pull/2110/commits/3bbe211c71e92e0c7b58b5b8f34d8aa946b796b3)

Akamig (approved):

src/renderer/views/LobbyView/index.tsx
```diff
@@ -33,15 +24,15 @@ function LobbyView() {
...
+  }, [history, error]);
 
   return (
     <>
       <Layout />
       <OnboardingOverlay
         // isOpen={onboardingRequired && showOnboarding}
         isOpen={false}
-        onClose={() => setShowOnboarding(false)}
+        onClose={() => {}}
```
longfin, line 35: Is this handler still needed?

longfin (approved):

src/renderer/views/LobbyView/index.tsx
maxswjeon: I don't think so, but the `onClose` prop was a required prop.

longfin: Then... how about loosening `OnboardingOverlay.onClose`'s requirement as optional?

maxswjeon: I'll check and loose the restraint. Thanks for your review!

Akamig: Yup, that OnboardingOverlay supposed to be a welcome message to newly activated user, but never actually used properly. 

Since we now have that as a subview in this PR, I can safely confirm you can remove that.👌

PR #2112, opened 2023-03-07
Zero-pad on 'short-key' scenario and rewrite keyfile.

### LINKED ISSUE #2107: Regenerate "short" Web3 Key to be more safer.
We need to at least zero-pad RawKey shorter than 32 byte and re-wrap into Web3 Secret Storage again.
Main issue on this is,

- Detecting if the raw key is shorter than 32 byte.
- Preventing re-padding already zero-padded key.
- Decide key regeneration behaviour, whether to rewrite Web3 key with same filename, or create new key file and deleting old key file.

Akamig:
This resolves #2107.

It's..... somewhat wacky code but works safely as my tests.

I executed following test.

- Web3 SS with key length 32 -> Passed login logic (Normal scenario)
- Web3 SS with key length 1 -> Padded rewrite success
- Web3 SS with key length 32 (padded from 1) -> Passed login logic

Key length 'longer than' 32, zero, negative length is not covered, but I guess that's pretty much out of this PR's scope.


src/stores/account.ts
```diff
@@ -245,4 +250,23 @@ export default class AccountStore implements IAccountStore {
...
+        v.path,
+        (
+          await new Wallet(
+            new SigningKey(
+              Buffer.concat([Buffer.alloc(32 - key.length), key], 32)
+            )
+          ).encrypt(passphrase)
+        )
+          .replace(/"id":"[0-9A-z-]*"/, `"id":"${v.keyId}"`)
+          .replace("C", "c")
```
Akamig, line 267: This was added due to encrypt method of `ethers` is generating json with uppercase "Crypto", which should be lowercase.

src/stores/account.ts
```diff
@@ -245,4 +250,23 @@ export default class AccountStore implements IAccountStore {
     this._privateKeyToRecovery = null;
     return account;
   };
+
+  fixKeyLength = async (v: ProtectedPrivateKey, passphrase: string) => {
+    const key = V3toRaw(await fs.promises.readFile(v.path, "utf8"), passphrase);
+    if (key.length < 32) {
+      fs.promises.writeFile(
```
longfin, line 257: It would be better to unify the style with `await` instead of callback..

src/stores/account.ts
```diff
@@ -245,4 +250,23 @@ export default class AccountStore implements IAccountStore {
...
+          await new Wallet(
+            new SigningKey(
+              Buffer.concat([Buffer.alloc(32 - key.length), key], 32)
+            )
+          ).encrypt(passphrase)
+        )
+          .replace(/"id":"[0-9A-z-]*"/, `"id":"${v.keyId}"`)
+          .replace("C", "c")
+      );
+      key.fill(0);
```
longfin, line 269: is this still needed? (because `key` variable doesn't seem to be used furthermore)

src/stores/account.ts
```diff
@@ -245,4 +250,23 @@ export default class AccountStore implements IAccountStore {
     this._privateKeyToRecovery = null;
     return account;
   };
+
+  fixKeyLength = async (v: ProtectedPrivateKey, passphrase: string) => {
```
longfin, line 254: I believe that leaving comments here to guess the context(i.e., what is the "fix"), for future readers would be nice.

longfin (commented):
Before testing, I left comments first. 🙏

src/stores/account.ts
Akamig: This effectively wipes raw private key regardless of garbage collection time.
This was previously questioned that this might cause copy-on-write and does not modify physical data on memory,

But I checked Node.js and ECMAScript specification thoroughly, and concluded
Buffer - Uint8Array - [ArrayBuffer](https://tc39.es/ecma262/multipage/structured-data.html#sec-arraybuffer-objects) - [DataBlock](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-data-blocks) is a mutable data block, and doesn't cause copy-on-write implicitly.

So yep, I think it's needed.

src/stores/account.ts
Akamig: Hmm... that what could be the better way? that;
1. Change function name to `padShortKey` and leave no comment
2. Comment what it does.

src/stores/account.ts
Akamig: Um... I'm kinda not getting it. you mean entire account.ts should be unify style with await, or this function specifically? 

I used `.then()` on some function intentionally (getAccount() for example) to bit more easily use that without any additional async handling.

longfin: 1. 
I think that `async`/`await` style is more helpful than .then() chaining.

> I used .then() on some function intentionally (getAccount() for example) to bit more easily use that without any additional async handling.

If you give an example in which case it is useful to use `then()` chaining, I think we can think together. 🤔 

2. But even if I support the async/await style, I also don't think it's important to fix it all at once considering the size of the PR. it is a comment that does not increase further within the range of change. 😅

src/stores/account.ts
longfin: - I think that if some volatile process can look up the physical memory of another process, there're many things we should consider and... I guess it doesn't seem practical. [^1]
- Of course, it's not a serious trade-off, so cleaning up memory isn't necessarily something to be avoided here. But it seems necessary to leave a comment about it.

[^1]: The reason [Spectre](https://en.wikipedia.org/wiki/Spectre_(security_vulnerability)) is a serious security vulnerability is that it can refer to the memory of other processes, which means that it is an important assumption of modern software development that you cannot see it.

src/stores/account.ts
longfin: `padShortKey` seems nice. but I guess comments might be still useful because it's too complicate and odd situation to cover naming only. 😅

src/stores/account.ts
Akamig: Left a comment on two different case of zero-filling.

src/stores/account.ts
longfin: To represent intention, how about replacing `"Crypto"` to `"crypto"`?

src/stores/account.ts
Akamig: Oh, my bad. I just assumed getAccount() must be sync function, which is not. I can fix this.

Akamig:
I guess review is pretty much applied at this point.

src/stores/account.ts
```diff
@@ -85,10 +89,14 @@ export default class AccountStore implements IAccountStore {
...
-  getAccount = (address: Address, passphrase: string): Promise<Account> => {
-    return this.findKeyByAddress(address).then((key) => {
-      return getAccountFromFile(key.keyId, passphrase);
-    });
+  getAccount = async (
+    address: Address,
+    passphrase: string
+  ): Promise<Account> => {
+    const key = await this.findKeyByAddress(address);
+    // Short Key Pad Check Flow
```
longfin, line 97:
```suggestion
    // Padding the existing key with the short length, created by Libplanet's previous bug.
    // see also: https://github.com/planetarium/9c-launcher/issues/2107
```

src/stores/account.ts
```diff
@@ -241,8 +247,35 @@ export default class AccountStore implements IAccountStore {
...
+          await new Wallet(
+            new SigningKey(
+              Buffer.concat([Buffer.alloc(32 - key.length), key], 32)
+            )
+          ).encrypt(passphrase)
+        )
+          .replace(/"id":"[0-9A-z-]*"/, `"id":"${v.keyId}"`)
+          .replace("C", "c")
+      );
+      // Wipe Buffer
```
longfin, line 277:
```suggestion
      // Wipe Buffer for security concerns.
      // See also: https://github.com/planetarium/9c-launcher/pull/2112#discussion_r1128982145
```

src/stores/account.ts
```diff
@@ -241,8 +247,35 @@ export default class AccountStore implements IAccountStore {
...
+        v.path,
+        (
+          await new Wallet(
+            new SigningKey(
+              Buffer.concat([Buffer.alloc(32 - key.length), key], 32)
+            )
+          ).encrypt(passphrase)
+        )
+          .replace(/"id":"[0-9A-z-]*"/, `"id":"${v.keyId}"`)
+          .replace("C", "c")
```
Akamig, line 275: Alright.
```suggestion
          .replace("Crypto", "crypto")
```

longfin (approved):
🙇

PR #2114, opened 2023-03-09
fix: activation flow

maxswjeon:
Fixed
- Loading Icon
- Text Update

Not fixed
- Validation
   The regex used for activation key check is `/^[0-9a-f]+\/[0-9a-f]{40}$/`, but it does not exclude private key validation.
   Need to use more strict activation key check.


Akamig:
>The regex used for activation key check is /^[0-9a-f]+\/[0-9a-f]{40}$/, but it does not exclude private key validation.
Need to use more strict activation key check.

Um... Unfortunately our activation key which consists of privatekey/address pair, the latter doesn't addresses former, instead, using activationKeyNonce for generating that for some reason, so basically in order to 'properly' validate if this is proper activation key, that we need to query the nonce value and do... some weird stuff. I did try once, but seems not worth the effort much, since mostly users are going to be copypasting.

https://github.com/planetarium/lib9c/blob/v100361/Lib9c/Model/State/PendingActivationState.cs#L31

Akamig (approved):
Except that, this LGTM!

maxswjeon:
Just for clarification of the bug that occured in the internal QA. (cc: @Gilhwan)
onChange validation works, but the difference between auto-paste (focus handler) and the manual-paste was the cursor location, resulting the change of the activation key location differ.

For more easy understanding, let's say `|` is the cursor location.
- Auto-paste makes the cursor location here.
   |abcedfgef/someaddresshere
- Manual-paste makes the cursor location here.
   abcdefgef/someaddresshere|
resulting the person who is in QA to make a change
|abc _d_ efgef/someaddresshere when auto-pasted, and
`abcdefgef/somea _d_ dresshere when manual-pasted

Since the regex does not check the length on the left side of the `/` (the private side of the activation key), it passes the naive activation key test

maxswjeon:
Some big changes:
- Ditched `Subview`, renamed to `View` and provided each routes per view.
- Removed `MissingActivationView`, since it is not required after the update.

longfin (approved):

PR #2129, opened 2023-04-07
[Feature] Migrate circleci to action

Atralupus:
Related to https://github.com/planetarium/9c-launcher/issues/2132

Migrate CircleCI to GitHub Actions for later enhancements in the launcher's CI/CD process
please check this action (https://github.com/planetarium/9c-launcher/actions/runs/4677166441)

PR #2133, opened 2023-04-13
Install dependencies with `--immutable` option

moreal:
Since #2129, Circle CI workflow was migrated into GitHub Actions workflow. During the steps, I found something was missing. This pull request recovers the `--immutable` option when installing dependencies with the yarn command.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,9 @@ jobs:
 
       - name: Install
         shell: bash
-        run: yarn
+        run: yarn install --immutable
+        env:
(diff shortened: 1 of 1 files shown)


maxswjeon:
I think it is more convenient to enable immutable option by environment variable, using `YARN_ENABLE_IMMUTABLE_INSTALLS`. I agree that `--immutable` option is more recommended, but there are cases where we should disable immutable installs.

Atralupus:
sorry, please rebase

Akamig (approved):

PR #2134, opened 2023-04-14
[Fix] Check disk space during Player/Launcher update

### LINKED ISSUE #1679: Check remaining disk size before updating
Currently, the launcher tries to update even if the user has not had enough space to download the newer version, which causes the launcher to be stuck on the update screen.

We could check the binary size of the update before downloading and compare the disk size against it, like how we did it with the snapshots.

upa-r-upa:
Resolve #1679 

In this PR, when a user is unable to update the launcher/player due to insufficient disk space, instead of simply stopping the launcher as before, now the launcher can guide the user through the error situation without stopping.

Initially, I wanted to proceed by checking the disk space before downloading and not even starting the download if there wasn't enough space. However, it was difficult to know the size of the uncompressed file before unpacking the downloaded file. Therefore, I opted to redirect the user to an error page to handle errors during the download process.

To create the error page, I couldn't find an appropriate existing page, so I created a page called `download-binary-failed-disk-error`. This page currently needs to be localized.
image

The three main changes from the previous implementation are as follows:

- Creation of an error page indicating that the update cannot proceed due to insufficient disk space
- In the electron-dl's onStarted event for downloading the Zip file, the remaining disk space is compared to the total file size, and if there is not enough space, the download is stopped and the user is redirected to the error page.
- If there is an error when unpacking the Zip file, the user is redirected to the disk space error page.

Please let me know if I misunderstood the requirements. Thank you.

### DIFF
--- a/src/main/update/player-update.ts
+++ b/src/main/update/player-update.ts
@@ -1,6 +1,7 @@
 import { DownloadItem, app } from "electron";
 import { download, Options as ElectronDLOptions } from "electron-dl";
 import { IDownloadProgress } from "src/interfaces/ipc";
+import { exec } from "child_process";
 import { DownloadBinaryFailedError } from "../exceptions/download-binary-failed";
 import fs from "fs";
 import extractZip from "extract-zip";
@@ -9,6 +10,34 @@ import { IUpdate } from "./check";
 import { configStore, playerPath, PLAYER_METAFILE_VERSION } from "src/config";
 import { createVersion } from "./metafile";
 
+type OSPlatform = "darwin" | "linux" | "win32";
+
+function getAvailableDiskSpace(path: string = "/"): Promise<number> {
+  return new Promise((resolve, reject) => {
+    const platform = process.platform as OSPlatform;
+    let cmd: string;
+    if (platform === "darwin" || platform === "linux") {
+      cmd = `df -k ${path} | tail -1 | awk '{print $4}'`;
+    } else if (platform === "win32") {
+      cmd = `for /f "skip=1" %p in ('wmic logicaldisk get FreeSpace^,Size^,Caption /format:list ^| find /i "${path}"') do @echo %p`;
+    } else {
+      return reject(new Error(`Unsupported platform: ${platform}`));
+    }
+
+    exec(cmd, (err, stdout, stderr) => {
+      if (err) {
+        return reject(err);
+      }
+      if (stderr) {
+        return reject(new Error(stderr));
+      }
+
+      const available = parseInt(stdout.trim(), 10);
+      resolve(available);
+    });
+  });
+}
+
 export async function playerUpdate(
   update: IUpdate,
   win: Electron.BrowserWindow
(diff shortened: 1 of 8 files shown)


src/utils/file.ts
```diff
@@ -0,0 +1,29 @@
+import { exec } from "child_process";
+
+type OSPlatform = "darwin" | "linux" | "win32";
+
+export function getAvailableDiskSpace(path: string = "/"): Promise<number> {
+  return new Promise((resolve, reject) => {
+    const platform = process.platform as OSPlatform;
+    let cmd: string;
+    if (platform === "darwin" || platform === "linux") {
```
Atralupus, line 9: Recommend dividing this 'if' condition into separate parts for better readability

Akamig (approved):
So if I summarize, this code would:
1. check size pre-download on ALL OS
2. check size during extraction of zip on WINDOWS
3. no specified size check during extraction on NON WINDOWS, but I think it would be handled at System call - operating system level. so seems okay for me.

Looooooooks Goooooooooood Tooooooooooooo Meeeeeeeeeeeeee

Ｔ ｈ ａ ｎ ｋ 
ｙ ｏ ｕ Ｆ ｏ 
ｒ ｙ ｏ ｕ ｒ
Ｃ ｏ ｎ ｔ ｒ 
ｉ ｂ ｕ ｔ ｉ 
ｏ ｎ ．

upa-r-upa:
> So if I summarize, this code would:
> 
> 1. check size pre-download on ALL OS
> 2. check size during extraction of zip on WINDOWS
> 3. no specified size check during extraction on NON WINDOWS, but I think it would be handled at System call - operating system level. so seems okay for me.
> 
> Looooooooks Goooooooooood Tooooooooooooo Meeeeeeeeeeeeee
> 
> Ｔ ｈ ａ ｎ ｋ ｙ ｏ ｕ Ｆ ｏ ｒ ｙ ｏ ｕ ｒ Ｃ ｏ ｎ ｔ ｒ ｉ ｂ ｕ ｔ ｉ ｏ ｎ ．

Oh no, I'm sorry. 😢 
The test code was accidentally included in the PR.

```typescript
 win.webContents.send("go to error page", "player", {
        size: e.size,
        url: "download-binary-failed-disk-error",
      });
```
This part was not accidentally deleted.

In cases where the compression fails due to lack of disk space, errors occur without being able to accurately determine the file size on both Windows, Linux, and Mac systems. I apologize for the confusion.

The case where the error page displays the size is only possible when downloading a zip file.

I have now made the necessary changes and pushed the updated code. Thank you!

src/utils/file.ts
upa-r-upa: Thank you for the feedback. I have made the necessary changes as per your suggestions!

src/main/update/player-update.ts
```diff
@@ -16,9 +17,24 @@ export async function playerUpdate(
   console.log("Start player update", update.projects.player);
   win.webContents.send("update player download started");
 
+  const available = await getAvailableDiskSpace(app.getPath("temp"));
+
   // TODO: It would be nice to have a continuous download feature.
   const options: ElectronDLOptions = {
     onStarted: (downloadItem: DownloadItem) => {
+      const totalBytes = downloadItem.getTotalBytes();
+      const totalKB = totalBytes / 1024;
```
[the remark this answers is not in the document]
Atralupus: While the zipped version of our player's file has a size of 400MB, the unzipped size amounts to 1.4GB.
It seems that issues occurring during the extraction process are handled elsewhere, so our primary focus should be on examining the downloaded file. 😵‍💫

Atralupus (approved):

PR #2137, opened 2023-04-17
feat:monster collection ui update

maxswjeon:
This PR introduces dynamic width until 9-digit numbers, and more levels


maxswjeon:
There is a missing asset right now, please don't merge it before the asset is added.

Akamig (approved):

PR #2155, opened 2023-05-04
Enhancement updater stability

Atralupus:
- Add confirm update view
- Enhancement stability

### DIFF
--- a/src/worker/checkForUpdateWorker.js
+++ b/src/worker/checkForUpdateWorker.js
@@ -23,37 +23,47 @@ setInterval(() => {
   checkForLauncherUpdate();
 }, 60 * 5 * 1000);
 
-function checkForPlayerUpdate() {
+async function checkForPlayerUpdate() {
   sendLog("debug", "Check for player update");
 
-  https
-    .get(`${baseUrl}/player/latest.json`, (response) => {
-      let data = "";
+  let response;
+  try {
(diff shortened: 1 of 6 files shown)


Akamig (approved):
appropriate timespan for update checking, other part checked, LGTM.

src/worker/checkForUpdateWorker.js
```diff
@@ -24,36 +38,71 @@ setInterval(() => {
...
-              if (file.os === os) {
-                process.send?.({
-                  type: "player update",
-                  path: `${baseUrl}/player/${file.path}`,
-                  size: file.size,
-                });
-              }
-            }
-          }
+          const local = readLocalPlayerVersionFile(err, playerVersionData);
```
longfin, line 41: It also would be better if we can use [`util.promisify()`](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original) for `fs.readFile()`, to replace the callback chain to promise based control flow.

longfin (approved):

src/worker/checkForUpdateWorker.js
Atralupus: Thank you. It may be less stable, but I missed it.
I have completed the task and would appreciate your review.

Atralupus: I had received approval, so I proceeded to merge the PR. 👀

PR #2163, opened 2023-05-18
Use lib9c-wasm on generating ActionTx used in launcher

Akamig:


Akamig:
Since there is a lot of diff forward this commit, 
and a lot of future changes are expected on both lib9c-wasm and this project too, I'm closing this issue.

PR #2203, opened 2023-06-13
Add Keystore Datamatrix Print Overlay (Export)

### LINKED ISSUE #2202: Datamatrix Export View

Akamig:
Resolves #2202

(Still working on UI)


mead.config.json
```diff
@@ -0,0 +1,47 @@
+{
```
longfin, line 1: This file doesn't seem to be needed for #2202

bitgener.d.ts
```diff
@@ -0,0 +1,63 @@
+export = bitgener;
```
longfin, line 1: it might be better than publishing this type as a separate package.

Or contributing upstream can be an option.

Akamig: I talked to maintainer and It seems they want to keep their library totally pure javascript. so....

mead.config.json
Akamig: Oh, That was my fault.

bitgener.d.ts
longfin: Then, how about publishing on [DT](https://definitelytyped.github.io/)?

Akamig: Alright!

24pAs (approved):

Akamig:
We should hold this PR until concrete representation approved by publishing team.

PR #2205, opened 2023-06-16
Update GQL Schema & Remove Unused Subscription API/Hooks

Akamig:
- Updated `schema.graphql` up to current main nodes
- Removed deleted monster collection related subscriptions.
- Also removed unused hook that uses that subscription queries.


Akamig:
Checked there was no components using that obsoleted hook (useMonsterCollection, openMonsterCollection), confirmed build success, merging.

Issue #2165, opened 2023-05-21
webpack file-loader plugin prevents us from using Node 18+


Akamig:
In short, `file-loader`
- is Webpack plugin we're using to bundle svg/jpg/png files,
- has been deprecated after Webpack v4,
- And if building environment is Node 16+, it won't build with following message : `Error: error:0308010C:digital envelope routines::unsupported`
- You can find about this issue on https://github.com/facebook/create-react-app/issues/12977
- Webpack officially encourages to migrate to [asset modules.](https://v4.webpack.js.org/loaders/file-loader/)
- So I think we should do that.

PR #2206, opened 2023-06-16
General... Dependancy Version Updating... and Cleaning...

Akamig:
BroomCloset

> Stanley stepped into the broom closet, but there was nothing here, so he turned around and got back on track.
> 
> There was nothing here. No choice to make, no path to follow, just an empty broom closet. No reason to still be here.
> 
> It was baffling that Stanley was still just sitting around in the broom closet. He wasn't even doing anything, at least if there were something to interact with he'd be justified in some way.

This is broom closet.


Akamig:
Update:
I've been trying to upgrade and remove unused packages, while resolving #2165, 
which is already included in this pr, but currently it's failing to build for some reason.

Akamig:
Estoy muy cansada ahora

Akamig:
I think it's time to merge this and release it at lease on 2.3.7.
This was always bugging in my mind. and stylebook still needs to fix, 
But we can do that separately.  

Alas! Mr. Stanley.
https://github.com/planetarium/9c-launcher/assets/6278999/69723403-aa65-434a-830e-582c90941fc1

Issue #2182, opened 2023-06-05
Apple Code Signing


Atralupus:
https://www.electron.build/code-signing

We have added a Mac updater in the merger after [v200020](https://github.com/planetarium/9c-launcher/pull/2173).
However, we can't use the auto updater because we didn't sign the binary. Refer to this [Document](https://www.electron.build/auto-update) for more details.
To fix this, we just need to add a code signing step in the Github action.

PR #2217, opened 2023-06-26
[Feature] Introduce apple code signing and notarization

Atralupus:
Close https://github.com/planetarium/9c-launcher/issues/2182

Implement apple code signing and notarization in ci/cd


signing/entitlements.inherit.plist
Akamig, line 1: welp, that's enough for our service I guess?
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/templates/entitlements.mac.plist

Akamig (approved):

Issue #2236, opened 2023-07-17
Print Current Launcher Version in Settings Overlay


Akamig:
I think it would be nice to acknowledge what version of launcher is running from user interface, 

- Helps backtracking issue at the certain point of time (between versions)
- Easily distinguishable if update condition is properly met or not,
- Confirms which launcher currently installed (internal and main-net)

nvsai:
Hi @Akamig I would like to try this issue could you please assign me?

Akamig:
@nvsai 
OH YES
THANKS
WELCOME
MUCH APPRECIATED

You can create fork, or branch from current `development` branch and work from there!
As saying "version", it's semver from package.json. (currently 2.1.1)
you don't have to adjust that to 2.1.2 after your contribution, but anyway.

If you have any problem while working, ask anything to me!

nvsai:
Hi @Akamig Thank you for assigning me this issue.
![Screenshot from 2023-07-18 11-08-45](https://github.com/planetarium/9c-launcher/assets/65706186/add31f61-fad0-45ca-b8f1-9e17eb494aa1)

I have placed the version in left bottom.And i have made changes in webpack.config to export version from package.json.Please let me know if any changes are required.

Akamig:
@nvsai 
Oh, no. Everything is perfect! "click-to-paste" is also nice touch!
Thanks for your contribution. 
Create PR, then I'll review your fix very quickly.

nvsai:
@AkamigI have created the PR.This would be my first open source contribution for a project of this scale Thank you very much.

PR #2240, opened 2023-07-18
Added Launcher version number in settings overlay.,

nvsai:
Print Current Launcher Version in Settings Overlay


Akamig (approved):
Confirm tested on my local environment, Working.

You did a great job! I just thought it would only paste the version, But also commit hash?! I haven't thought that, but we specifically needed that during internal testing! You really out did yourself. Thanks for everything. Have a nice day. 🥰

Akamig:
Oh, I forgot to mention, as you said this is first contribution for this scale of project,
This is a little tip for you! 

When you making PR, if you mentioning number of resolving issue something like
"This resolves #2236."
That issue automatically associated with this PR, and automatically closes that issue when PR merged.

src/renderer/views/SettingsOverlay/index.tsx
```diff
@@ -108,7 +111,24 @@ async function handlePlayerUpdate() {
...
+  bottom: 50,
+  left: 50,
+  dragable: false,
+});
+
 function SettingsOverlay({ onClose, isOpen }: OverlayProps) {
+  let versionToDisplay = "unknown";
+  try {
+    versionToDisplay = CURRENT_VERSION;
+  } catch (error) {
```
longfin, line 128: this `catch` clause seems to be a little bit weird since the `try` block has only one assigning statement. 😕 

can we elide try-catch blocks? (or replacing it with a`CURRENT_VERSION` check might be also good).

src/renderer/views/SettingsOverlay/index.tsx
```diff
@@ -7,8 +7,9 @@ import { configStore, userConfigStore } from "src/config";
 import type { IConfig } from "src/interfaces/config";
 import { T } from "src/renderer/i18n";
 import log from "electron-log";
-import { shell, ipcRenderer } from "electron";
+import { clipboard,shell, ipcRenderer } from "electron";
```
longfin, line 10:
```suggestion
import { clipboard, shell, ipcRenderer } from "electron";
```

src/renderer/views/SettingsOverlay/index.tsx
```diff
@@ -147,6 +167,27 @@ function SettingsOverlay({ onClose, isOpen }: OverlayProps) {
     []
   );
 
+  const address = useLoginSession()?.address;
+
+  const debugValue = useMemo(
+    () =>
+      [
+        `LAUNCHER_VERSION: ${versionToDisplay}`,
+        address && `Account: ${address.toString()}`,
```
longfin, line 176:
```suggestion
        address && `Account: ${address}`,
```

In TypeScript, the embedded variable in the template string will be applied as `.toString()` automatically. 😅

Akamig: Ah, but I prefer using `.toString()` in this case, because while just print `address` doesn't, using `.toString()` automatically attaches `0x` prefix. which is preferred.

longfin: AFIK, `Account: ${address.toString()}` and `address && `Account: ${address}` will produce same results.

src/renderer/views/SettingsOverlay/index.tsx
Akamig: Hmm... well, I thought this is relatively safer for "just in case", since pre-declared `CURRENT_VERSION` const is not nullable and build process itself would _probably_ break if that's the case, so basically it would filter in case build success and somehow working while fail to fetch proper version. 

Actually I might add some more other validation logic in that `try`. as far as I remember we have some version-switcheroo logic in internal target CI.

src/renderer/views/SettingsOverlay/index.tsx
Akamig: I just tested. and welp, you are right. 
I was confused with `address.toHex()` which was the not prefix giving one.

...but while that, I still think `toString()` should be used for just for preventing future developers get confused about this prefix attaching behavior is default or not. what do you think?

longfin: Basically, I don't think that `.toString()` usage would help with confusion about 0x prefix because this confusion has been related to `.toString()` and `.toHex()` semantics, not about string interpolation.

PR #2261, opened 2023-07-25
Hide to tray on click X, quit on tray

### LINKED ISSUE #2180: V20020 - Launcher closes instead of sending it to tray when pressing X
Hi Team,

Not sure if it is a bug or intentional, however with the launcher that's on V20020 when you press X on the UI which previously sent it to system tray, it now closes the whole launcher, which then subsequently takes down the 9C.exe Client with it.

If it was done intentionally, could we considering making an extra option to send it to the tray? From what I could see in the community chat, everyone expects it to go back to the tray.

Akamig:
This resolves #2180, 
image
- Add "Keep launcher on tray" when closed option on Settings (default: true)
- Optionally keep launcher opened at tray when closed
- Tray - Quit Launcher to really close
- Fix launcher updater not affected by optional hide cc @Atralupus 

Tested on mac, working flawlessly.

~~This resolves #2180 in different way, instead of push close to minimize to tray, minimize button simply hide launcher into tray.~~

~~Um, it doesn't work on macOS as I tested? going to test on windows.~~

~~Sorry for late fix, @FioX0.~~

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -227,6 +227,7 @@ async function initializeConfig() {
     // Replace config
     console.log("Replace config with remote config:", remoteConfig);
     remoteConfig.Locale = getConfig("Locale");
+    remoteConfig.TrayOnClose = getConfig("TrayOnClose", true);
     console.log(remoteConfig.Locale);
     configStore.store = remoteConfig;
   } catch (error) {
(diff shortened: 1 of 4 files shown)


src/main/main.ts
```diff
@@ -359,6 +362,7 @@ function initializeIpc() {
 
   ipcMain.handle("execute launcher update", async (event) => {
     if (appUpdaterInstance === null) throw Error("appUpdaterInstance is null");
+    setV2Quitting(true);
```
Akamig, line 365: This is the part that keep quitAndInstall in appUpdaterInstance quit properly,
and I think this is the only entry to trigger quitAndInstall directly. 
but if I'm missing anything. please nudge me.

Atralupus (commented):
I'm not certain whether this code will function properly due to my limited understanding of electron.
However, it seems fine to me.

PR #2267, opened 2023-07-28
feat: Upgrade apollo client version

Atralupus:
to 3.7.17


Akamig:
I haven't seen any noticeable breaking change from their API from this update, so I think it's good to go.

PR #2280, opened 2023-08-09
Display FAV in MonsterCollection Reward

### LINKED ISSUE #2279: Display FAV in Monster Collection Reward
Currently, There's no identifier for check Currency type of reward. so we simply omit them to print.

But since we're planning to add more FAV based reward, 
1. We need to print them properly
2. But in order to do that we need identifier like ItemID
3. Add CurrencyTicker as query-able field -> Headless PR

Akamig:
# DO NOT MERGE THIS UNTIL NEW HEADLESS DISTRIBUTED
This resolves #2279, but this code in effect after 
https://github.com/planetarium/NineChronicles.Headless/pull/2185
resolved. so do not merge this PR until that sorts out.

Anyway, This PR adds identifier for each Currency tickers, and re-uses legacy display code.

### DIFF
--- a/src/api.graphql
+++ b/src/api.graphql
@@ -273,6 +273,7 @@ query StakingSheet {
           itemId
           rate
           type
+          currencyTicker
         }
         bonusRewards {
           itemId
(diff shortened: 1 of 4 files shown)


Akamig:
Tested on changed GraphQL Schema and adapted on former mentioned API.

PR #2323, opened 2023-09-20
Use the patronAddress from the pladge states

longfin:
This PR adds the argument for `approvePledge` on GQL API, to support other patrons than hardcoded one.

### DIFF
--- a/src/api.graphql
+++ b/src/api.graphql
@@ -392,8 +392,8 @@ query CheckContracted($agentAddress: Address!) {
   }
 }
 
-query approvePledge($publicKey: String!) {
+query approvePledge($publicKey: String!, $patronAddress: Address!) {
   actionTxQuery(publicKey: $publicKey) {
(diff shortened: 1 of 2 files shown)


Akamig:
@longfin 
Hmm... as far as I remember I did this fixed address because I was not sure how can I validate patron address before using it to send approvePledge. although it would be fail at chain level in that case, can we disregard the possibility of malicious patron at current status?

Akamig:
(Either way I agree with this flexible patron, just not sure it should goes like SwapAddress or else)

Akamig (approved):
> @longfin Hmm... as far as I remember I did this fixed address because I was not sure how can I validate patron address before using it to send approvePledge. although it would be fail at chain level in that case, can we disregard the possibility of malicious patron in long term?

Welp, since there's no possible method of malicious MEAD distribution at current status I can think of, I approve this.

PR #2351, opened 2023-10-26
Polish Select-Box

Akamig:
Screenshot 2023-10-26 at 3 46 47 PM


src/renderer/views/LoginView/index.tsx
```diff
@@ -93,35 +98,40 @@ function LoginView() {
...
-              {accountStore.addresses.map((address) => (
-                <SelectOption key={address.toHex()} value={address.toHex()}>
-                  {address.toString()}
-                </SelectOption>
-              ))}
-            </Select>
-          )}
-        />
+      <SelectWrapper fullWidth>
+        <SelectLabel id="planet-label">Planet</SelectLabel>
```
longfin, line 102: it would be better with `<T />` tag for i18n

PR #2370, opened 2023-11-10
Use bridge on interplanetary transactions.

Akamig:
Figure-3-13-Hohmann-transfer-technique-Starting-from-A-on-the-internal-circle-a

### DIFF
--- a/src/renderer/views/TransferAssetOverlay/main.tsx
+++ b/src/renderer/views/TransferAssetOverlay/main.tsx
@@ -87,7 +87,8 @@ const DescriptionTitleMessage = styled(Typography)({
 });
 
 function TransferAssetOverlay({ isOpen, onClose }: OverlayProps) {
-  const isOdin = useStore("planetary").planet.id === "0x000000000000";
+  const isOdin = useStore("planetary").planet.id.endsWith("000");
+  // This is only safe until 4095 other planet registered in theory, so be aware of it.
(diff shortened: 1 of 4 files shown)


src/renderer/views/TransferAssetOverlay/main.tsx
```diff
@@ -87,7 +87,8 @@ const DescriptionTitleMessage = styled(Typography)({
 });
 
 function TransferAssetOverlay({ isOpen, onClose }: OverlayProps) {
-  const isOdin = useStore("planetary").planet.id === "0x000000000000";
+  const isOdin = useStore("planetary").planet.id.endsWith("000");
+  // This is only safe until 4095 other planet registered in theory, so be aware of it.
```
longfin, line 91: I think that defining planet ids for odin-like planet explicitly would be better than implicit rule.

Akamig: I beg to differ, in short,
- We should be able to test out this "Hide swap" feature on internal test too.
- if we're intended to use same environment between main/internal, using same planet ID scheme for corresponding chain would be advised.
- if not, we need to use something better to adapt on such policy, instead of hardwired-value or values, preferably on registry level.
  - especially if we're intending to selectively provide (or does not) certain service per-chain.

longfin: I also didn't think that defining as array is best solution. but it was better than current implicit way at least.

PR #2398, opened 2024-01-16
add optional guild service args for player

ipdae:
- related from https://github.com/planetarium/NineChronicles/pull/3924


Akamig (approved):
Running test confirmed

PR #2405, opened 2024-01-19
Override guild arguments from planetRegistry

ipdae:
override player guild service arguments from planet registry


Akamig (approved):
Checked Logic working, Approve.

PR #2441, opened 2024-05-02
Fix: rollback swap field fee calculation

Akamig:
image


Junyong-Suh:
@Akamig hmm what's the [Chromatic](https://github.com/planetarium/9c-launcher/actions/runs/8920811435/job/24499607292?pr=2441) that failed?

